PR #2, opened 2026-02-19
ensure ~/.nullclaw exists

jackharrhy:
running `nullclaw onboard`, would result in just `error: FileNotFound`, no further logs

this pr ensures the parent configuration dir exists before trying to write child files to it

tested, after this change, `nullclaw onboard` creates what it needs to create properly

### DIFF
--- a/src/onboard.zig
+++ b/src/onboard.zig
@@ -433,7 +433,13 @@ pub fn runQuickSetup(allocator: std.mem.Allocator, api_key: ?[]const u8, provide
         cfg.default_model = defaultModelForProvider(cfg.default_provider);
     }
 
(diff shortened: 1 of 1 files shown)


DonPrus:
Minor notes (non-blocking):
  - `makeDirAbsolute(dirname(...))` only creates one level of parent directory — works fine for the default `~/.nullclaw/workspace`, but will fail for custom deep paths like `/a/b/c/workspace` where intermediate dirs don't exist.                     
  - There's already `cfg.ensureDirs()` that does the same thing — could call that instead of duplicating the dirname + makeDirAbsolute pattern in two places.

PR #53, opened 2026-02-21
fix: use musl based image for release stage

### LINKED ISSUE #51: Error directory
### Description

command: docker-compose --profile agent up -d

Image

### Expected behavior

_No response_

### Steps to reproduce

_No response_

### Version

_No response_

### OS

None

mporrato:
Align the release stage base image to the image used for the build.

Since the binary is linked against the sqlite3 library, the resulting binary also depends on the libc of the build stage, which in this case is musl and won't work when executed in a glibc environment, like the debian-based distroless base.
I've also added a bunch of additional packages to make the out of the box setup somewhat usable.

Should fix #51

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,11 @@ EOF
 
 RUN chown -R 65534:65534 /nullclaw-data
 
-# ── Stage 3: Production Runtime (Distroless) ─────────────────
-FROM gcr.io/distroless/cc-debian13:nonroot AS release
+# ── Stage 3: Production Runtime ──────────────────────────────
+FROM alpine:3.23 AS release
+
+RUN apk add --no-cache ca-certificates tzdata curl \
+    openssh-client git sqlite uv python3 github-cli
 
 COPY --from=builder /app/zig-out/bin/nullclaw /usr/local/bin/nullclaw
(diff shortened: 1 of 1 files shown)


DonPrus:
Hey @mporrato, thanks for diagnosing the musl/glibc mismatch — spot on!

I went ahead and made a minimal fix in #57: just swapping the runtime base to `alpine:3.23` with `ca-certificates` and `tzdata`. The binary is statically linked against everything it needs (including SQLite), so it runs fine without additional packages in the container.

I left out `sqlite`, `uv`, `python3`, `github-cli`, `git`, `curl`, `openssh-client` to keep the image small. If users need any of those for tool use (e.g. shell commands from the agent), they can extend the image:

```dockerfile
FROM nullclaw:latest
RUN apk add --no-cache git curl openssh-client
```

Or use a docker-compose override / init script depending on the deployment setup.

mporrato:
Thanks @DonPrus  for looking into this and taking action. The extra packages were more of a convenience for me to troubleshoot a couple of issues I'm having.
I suppose, if you still want to keep the distroless base, you could statically link musl in the build stage but it's probably a more involved solution.
Closing.

PR #63, opened 2026-02-22
feat(channels): add Signal channel via signal-cli HTTP daemon

### LINKED ISSUE #58: [Feature Request] Add Signal channel support via `signal-cli --http` JSON-RPC daemon
### Description

As per the title, would love to see channel support for Signal using [signal-cli](https://github.com/AsamK/signal-cli).

I personally added Signal support to both zeroclaw https://github.com/zeroclaw-labs/zeroclaw/pull/468 and ironclaw https://github.com/nearai/ironclaw/pull/271.

If you're accepting PRs from LLMs I can work on this one.

Disclaimer: whereas I know rust pretty good which makes it easier to review and guide the models, this will be my first endeavor in Zig.

### Motivation

_No response_

ibhagwan:
Implement Signal messaging channel that connects to a signal-cli daemon running in HTTP mode (signal-cli --http <port>). Follows the vtable-based channel pattern established by existing channels.

Architecture:
- SSE event stream for receiving messages at /api/v1/events
- JSON-RPC for sending messages/typing indicators at /api/v1/rpc
- curl subprocess for all HTTP (consistent with other channels)
- Manual JSON construction via json_util helpers

Features:
- Full envelope processing (data messages, group messages)
- Allowlist-based access control for users and groups (empty allowlist = deny all, secure by default)
- Story message filtering (ignore_stories, default true)
- Attachment placeholder support (ignore_attachments, default true)
- Wildcard (*) support in allowlists
- UUID normalization (uuid: prefix) in allowlists
- E.164 phone number validation
- Trailing slash stripping from base URL
- Message splitting at 4096 char limit
- Best-effort typing indicators
- Health check via /api/v1/check endpoint

Config (accounts wrapper pattern):
  channels.signal.accounts.<id>.http_url     (required)
  channels.signal.accounts.<id>.account      (required, E.164 phone)
  channels.signal.accounts.<id>.allowed_users
  channels.signal.accounts.<id>.allowed_groups
  channels.signal.accounts.<id>.ignore_attachments (default: true)
  channels.signal.accounts.<id>.ignore_stories     (default: true)

Files:
- src/channels/signal.zig: channel implementation + tests
- src/config_types.zig: SignalConfig struct + ChannelsConfig field
- src/config_parse.zig: JSON config parsing for Signal channel
- src/channels/root.zig: sub-module registration

Closes #58

### DIFF
--- a/src/channel_loop.zig
+++ b/src/channel_loop.zig
@@ -250,6 +250,7 @@ pub fn runTelegramLoop(
                 const err_msg: []const u8 = switch (err) {
                     error.CurlFailed, error.CurlReadError, error.CurlWaitError => "Network error. Please try again.",
                     error.ProviderDoesNotSupportVision => "The current provider does not support image input. Switch to a vision-capable provider or remove [IMAGE:] attachments.",
+                    error.NoResponseContent => "Model returned an empty response. Please retry or /new for a fresh session.",
                     error.OutOfMemory => "Out of memory.",
                     else => "An error occurred. Try again or /new for a fresh session.",
                 };
@@ -371,6 +372,7 @@ pub fn runSignalLoop(
                 const err_msg: []const u8 = switch (err) {
                     error.CurlFailed, error.CurlReadError, error.CurlWaitError => "Network error. Please try again.",
                     error.ProviderDoesNotSupportVision => "The current provider does not support image input.",
+                    error.NoResponseContent => "Model returned an empty response. Please try again.",
                     error.OutOfMemory => "Out of memory.",
                     else => "An error occurred. Try again.",
                 };
@@ -566,6 +568,7 @@ pub fn runMatrixLoop(
                 const err_msg: []const u8 = switch (err) {
                     error.CurlFailed, error.CurlReadError, error.CurlWaitError => "Network error. Please try again.",
                     error.ProviderDoesNotSupportVision => "The current provider does not support image input.",
+                    error.NoResponseContent => "Model returned an empty response. Please try again.",
                     error.OutOfMemory => "Out of memory.",
                     else => "An error occurred. Try again.",
                 };
--- a/src/daemon.zig
+++ b/src/daemon.zig
@@ -375,6 +375,7 @@ fn inboundDispatcherThread(
             const err_msg: []const u8 = switch (err) {
                 error.CurlFailed, error.CurlReadError, error.CurlWaitError => "Network error. Please try again.",
                 error.ProviderDoesNotSupportVision => "The current provider does not support image input.",
+                error.NoResponseContent => "Model returned an empty response. Please try again.",
                 error.OutOfMemory => "Out of memory.",
                 else => "An error occurred. Try again.",
             };
--- a/src/gateway.zig
+++ b/src/gateway.zig
@@ -1222,6 +1222,26 @@ pub fn sendTelegramReply(allocator: std.mem.Allocator, bot_token: []const u8, ch
     _ = curl_child.wait() catch {};
 }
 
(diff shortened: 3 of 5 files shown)


ibhagwan:
Let me know quarter you think @DonPrus, I’m thinking to change the config option to match telegram and option names in Openclaw, so instead of allowed_users -> allowed_from, etc.

ibhagwan:
Also, I’m not familiar with the codebase yet but if the model implemented this in the most efficient way ( currently) there may be some work to be done in the channel abstraction layer as there are way too many changes needed in files outside the channel folder.

For example, onboarding, status, the main loop - all should be automatically propagated after a channel is added (vs the current hardcoded approach).

ibhagwan:
Ty @DonPrus,  fantastic changes

ibhagwan:
@DonPrus, https://github.com/nullclaw/nullclaw/pull/63/changes/f0c43cff25941b42574daacd0ab116b965061b47

> allowed_groups→group_allow_from

I’m not sure about this, if I am to judge this by the openclaw config conventions group_allow_from are user accounts (E164 numbers) which are allowed to talk within configured/permitted groups, the original code (perhaps changes now) uses  `allowed_groups` to limit which groups the agent participates in (unless the group is in this list or the list is `* ` the agent will ignore all messages in the group.

Judging by the chosen name I believe you want to match this behavior to openclaw, in which case we need to modify the code.

I’m thinking it’s still valuable to decide which groups the agent is listening to at all and on top to have group_allow_from which is the “allowlist” of groups only.

ibhagwan:
https://github.com/nullclaw/nullclaw/pull/63/changes/9812db56176b8bb1b48fa210da916d94b6a11067, FYI, the DebugAllocator was segfaulting due to a race condition on debug build when running the daemon.

DonPrus:
@ibhagwan, thanks, I will double check every thing tomorrow and than I will be ready for merge

ibhagwan:
@DonPrus, when building debug mode with simply `zig build` I get a `panic: unreachable code` as soon as the first message arrives (consistently), I tracked down to sqllite.sig implStore, I’m still working on it unsure why this happens and if related to the signal code, this doesn’t happen with ReleaseSmall.

Per my understanding it’s because ReleaseSmall doesn’t panic for unreachable but the behavior will be undefined and can lead to some nasty voodoo bugs.

I don’t have any other channels configured, can you check with another channel (or signal if you have) and lmk if you get the panic too in a debug build?

I’ve been working on it for a while now suspecting the memory is being corrupted somewhere else causing the vtable call to panic.

DonPrus:
@ibhagwan, I added some tests, but couldnt reproduce it now, could you try it again?

ibhagwan:
> @ibhagwan, I added some tests, but couldnt reproduce it now, could you try it again?

Pulled latest changes, still crashing on the first signal message in debug mode, it's very consistent for me but the source of the memory corruption is ellusive, I'm still debugging it will keep you posted.

Have you tried with signal channel or other channels only? I'm suspecting something in the newly added signal channel code.

ibhagwan:
Found the issue @DonPrus, it was only happening on memory.backend=markdown due to workspace_dir ref being corrupted, testing and will commit the fix in a sec.

ibhagwan:
@DonPrus, https://github.com/nullclaw/nullclaw/pull/63/commits/2539cdb3688d4f8c34cbd28fd7452271c5e266fb

DonPrus:
Looks good for me, lets wait checks and I will merge it.

ibhagwan:
> Looks good for me, lets wait checks and I will merge it.

Amazign work tysm @DonPrus!

PR #75, opened 2026-02-23
fix(docker): run container as root for agent autonomy

streambinder:
## Summary

- Remove `USER 65534:65534` so the agent can install packages at runtime via `apk`,
  enabling autonomous self-progression (e.g. installing build tools, linters, etc.)
- Remove the now-unnecessary `chown -R 65534:65534` on the data directory
- Rename the build stage from `permissions` to `config` to reflect its actual purpose

## Rationale

The `nobody` user prevents the agent from running package managers inside the container,
which blocks it from installing tools it needs to complete tasks autonomously. The container
itself serves as the security boundary; nullclaw's own allowlist-based command policy and
workspace sandboxing handle access control at the application layer.

## Test plan

- [ ] Build the image: `docker build -t nullclaw .`
- [ ] Run the container and verify the agent can execute `apk add` to install packages
- [ ] Verify nullclaw starts and serves the gateway normally on port 3000

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,8 @@ Key extension points:
 - `src/runtime.zig` (`RuntimeAdapter`) — execution environments
 - `src/peripherals.zig` (`Peripheral`) — hardware boards (Arduino, STM32, RPi)
 
+Current scale: **151 source files, ~96K lines of code, 3,371 tests**.
+
 Build and test:
 
 ```bash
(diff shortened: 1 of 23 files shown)


DonPrus:
I pushed an update to keep the default image safe (non-root) and add your root setup as an explicit opt-in mode.

How to use root mode now:

```cmd
docker build --target release-root -t nullclaw:root .
docker run --rm -p 3000:3000 nullclaw:root

```

Default build stays non-root:

```cmd
docker build -t nullclaw .
docker run --rm -p 3000:3000 nullclaw

```
So we keep secure-by-default behavior, while still supporting your autonomous/root workflow when explicitly requested.

PR #89, opened 2026-02-23
fix(config): require explicit default_model, remove paid fallbacks

streambinder:
## Summary

- **Parse top-level `default_model`** from config JSON — was silently skipped by `config_parse.zig`
- **Change struct default** from `"anthropic/claude-sonnet-4"` to `null` — no more implicit paid model
- **Add `NoDefaultModel` validation error** — nullclaw now fails loudly at startup if no model is configured
- **Remove all 8 hardcoded paid-model `orelse` fallbacks** scattered across `main.zig`, `agent/root.zig`, `channel_loop.zig`, and `providers/helpers.zig` (three different Sonnet versions were used inconsistently)
- **Print configured models on boot** across all entry points (gateway, daemon, channel, agent REPL)
- **Add `printValidationError()` helper** for consistent error output across entry points

## Problem

Setting `"default_model"` at the top level of `config.json` had no effect. The parser skipped this field entirely, so the struct default (`anthropic/claude-sonnet-4`) was always used instead. On top of that, every downstream consumer had its own `orelse` with a hardcoded paid Anthropic model — three different Sonnet versions, inconsistently.

**This is a serious concern for first-time users.** Someone configuring a free model (e.g. `meta-llama/llama-3.3-70b-instruct:free` on OpenRouter) would have no indication that their config was being ignored. The application starts normally, responds normally, and nothing in the logs suggests anything is wrong. The only way to discover the issue is when the OpenRouter bill arrives — potentially days later and already significant (in my case, the bill reported ~$9 in unexpected charges in a single day, for 199 API calls routed to `anthropic/claude-4-sonnet` instead of their configured free model). For users who set up nullclaw and leave it running on a channel, this could accumulate silently for days or weeks before anyone checks the bill.

No application should silently route traffic to a paid API when the user explicitly configured a free one.

### No model configured → exits with error

```
$ ./zig-out/bin/nullclaw gateway
No default_model configured. Set it in ~/.nullclaw/config.json or run `nullclaw onboard`.
$ echo $?
1
```

### Model configured via top-level `default_model` → starts and prints model

```json
{
  "default_provider": "openrouter",
  "default_model": "meta-llama/llama-3.3-70b-instruct:free",
  "default_temperature": 0.7
}
```

```
$ ./zig-out/bin/nullclaw gateway
Gateway listening on 127.0.0.1:3000
  Model:    meta-llama/llama-3.3-70b-instruct:free
  Provider: openrouter
Gateway pairing code: 520322
```

### Model configured via nested `agents.defaults.model.primary` → also works

```json
{
  "default_provider": "openrouter",
  "agents": {"defaults": {"model": {"primary": "anthropic/claude-opus-4"}}},
  "default_temperature": 0.7
}
```

```
$ ./zig-out/bin/nullclaw gateway
Gateway listening on 127.0.0.1:3000
  Model:    anthropic/claude-opus-4
  Provider: openrouter
Gateway pairing code: 195801
```

## Files changed (10)

| File | What changed |
|---|---|
| `src/config.zig` | `default_model` default → `null`, `NoDefaultModel` validation, `printModelConfig()`, `printValidationError()` |
| `src/config_parse.zig` | Parse top-level `default_model` key (the primary bug fix) |
| `src/main.zig` | Validate config at startup in gateway/daemon/channel, use `.?` instead of `orelse` |
| `src/agent/cli.zig` | Validate config at startup, print model config in REPL mode |
| `src/agent/root.zig` | Replace `orelse "anthropic/claude-sonnet-4"` with `return error.NoDefaultModel` |
| `src/channel_loop.zig` | Replace `orelse "anthropic/claude-sonnet-4"` with log + return |
| `src/providers/helpers.zig` | Replace both `orelse "anthropic/claude-sonnet-4-5-20250929"` with `return error.NoDefaultModel` |
| `src/daemon.zig` | Print model config on daemon startup |
| `src/gateway.zig` | Print model config on gateway startup |
| `src/session.zig` | Set `default_model` in test helper (was relying on old non-null default) |

## Test plan

- [x] All 3020 existing tests pass (`zig build test`)
- [x] New tests: `validation rejects null default_model`, `json parse top-level default_model`, `agents.defaults.model.primary overrides top-level default_model`
- [x] Manual: gateway refuses to start without `default_model`
- [x] Manual: gateway starts and prints model with top-level config
- [x] Manual: gateway starts and prints model with nested config

### DIFF
--- a/src/gateway.zig
+++ b/src/gateway.zig
@@ -2087,7 +2087,7 @@ pub fn run(allocator: std.mem.Allocator, host: []const u8, port: u16, config_ptr
     const stdout = &bw.interface;
     try stdout.print("Gateway listening on {s}:{d}\n", .{ host, port });
     try stdout.flush();
-    if (config_opt) |*cfg| {
+    if (config_opt) |cfg| {
         cfg.printModelConfig();
     }
     if (state.pairing_guard) |*guard| {


DonPrus:
Thanks for flagging this.

I want to drop support for the legacy top-level keys to keep the config model aligned with OpenClaw and make migration behavior clearer. Going forward, we’ll treat those legacy keys as invalid and return an explicit validation error, with guidance to use `agents.defaults.model.primary` instead.

PR #95, opened 2026-02-24
feat(signal): implement streaming SSE for real-time message delivery

ibhagwan:
Replace curl-based polling with persistent HTTP streaming using std.http.Client.
- Add sse_client.zig with SseConnection struct for persistent connections
- Integrate streaming into SignalChannel.pollMessages() with reconnect logic
- Add exponential backoff retry (2s to 60s) for connection failures
- Buffer incomplete SSE events across read calls
- Cleanup SSE resources in vtableStop
- Add parseEvents tests for SSE format validation

Avoids Zig 0.15 namespace collision by keeping sse_client separate from modules that import root.zig. Messages now arrive instantly instead of waiting for curl timeout (previously 60s, now continuous streaming).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,10 @@ jobs:
             target: linux-aarch64
             zig_target: aarch64-linux-musl
             ext: ""
+          - os: ubuntu-latest
+            target: linux-riscv64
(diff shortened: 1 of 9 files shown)


ibhagwan:
Thank you so much @DonPrus for your dedication to this project

PR #103, opened 2026-02-24
fix(docker): add OCI source label for ghcr visibility

streambinder:
## Summary

- Adds `org.opencontainers.image.source` label to the Dockerfile's release stage

## Context

Adding `LABEL org.opencontainers.image.source=https://github.com/nullclaw/nullclaw` allows GitHub to link the container package to the repo, enabling visibility and permission inheritance from the repository settings.

## Test plan

- [ ] Verify the label is present in the built image via `docker inspect`
- [ ] After merge and publish, confirm the ghcr.io package is linked to the repo and publicly accessible


streambinder:
@DonPrus on top of this, please, also let's make sure the ghcr image is set as public. I can't fetch it :(

Issue #113, opened 2026-02-25
Add a custom provider config sample to the `config.example.json` file


XhstormR:
Currently, there is no specific configuration sample or guide on how to add a custom provider.

DonPrus:
Thanks for raising this. Here is a working custom-provider example you can use today.

```json
{
  "models": {
    "providers": {
      "myproxy": {
        "api_key": "YOUR_API_KEY",
        "base_url": "https://api.example.com/v1",
        "native_tools": true
      }
    }
  },
  "agents": {
    "defaults": {
      "model": {
        "primary": "myproxy/my-model-name"
      }
    }
  }
}
```

How to use it:
1. Add the provider entry under `models.providers`.
2. Set `base_url` to your OpenAI-compatible endpoint.
3. Set the default model as `provider_key/model_name` (example: `myproxy/my-model-name`).

Notes:
- This custom-provider path works, but I have not deeply stress-tested every custom endpoint variant yet.
- I will add more advanced examples to `config.example.json` later (multiple providers, fallbacks, and routing patterns).

ariefsn-tsrlabs:
any idea how to add custom headers?

PR #119, opened 2026-02-25
fix(providers): respect base_url and add native_tools toggle for custom providers

sleep3r:
**Problem**

1. Unknown providers ignore base_url: Providers not in the built-in list (e.g. "local") silently fall back to OpenRouter, discarding the configured base_url.

2. Native tool calls break local proxies: The compatible provider always advertises native OpenAI tool_calls support, but nullclaw stores tool history as XML. This mismatch causes AllProvidersFailed errors when the proxy enforces strict OpenAI tool protocol.

**Changes**

1. factory.fromConfig: Unknown providers with base_url now create an OpenAiCompatibleProvider instead of falling back to OpenRouter. Config base_url also overrides built-in URL table for known compatible providers.

2. native_tools config flag: New per-provider boolean (default true). When false, the agent uses XML tool format via system prompt instead of structured tool_calls in the API request.

**Config example**
```
"local": {
    "base_url": "http://127.0.0.1:8045/v1",
    "api_key": "...",
    "native_tools": false
}
```

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -153,6 +153,15 @@ pub const Config = struct {
         return null;
     }
 
+    /// Look up whether a provider supports native tool calls.
+    /// Returns true (default) if provider is not in the list.
+    pub fn getProviderNativeTools(self: *const Config, name: []const u8) bool {
+        for (self.providers) |e| {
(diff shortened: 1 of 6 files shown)


sleep3r:
see #113

Issue #111, opened 2026-02-24
Typing indicator disappears after 10 seconds in Discord


vaneistech:
### Description

Thank you for implementing the visual confirmation when Nullclaw is thinking or replaying in Discord and other channels. 

I noticed a bug in the current implementation. The "... typing" indicator disappears after only 10 seconds. This causes confusion because the user cannot tell if the agent has crashed or stop processing a long request. 

In the original OC implementation, the indicator stays visible for the entire duration of the process. It would be great if you could adjust this so the indicator remains until the agent finishes its task. 

Thanks! 

### OS

None

PR #124, opened 2026-02-26
fix(signal): keep typing indicator alive for the full duration of processing

ibhagwan:
Signal's typing indicator expires shortly after being sent (similar to Discord). The previous implementation fired sendTypingIndicator once per inbound message, so the indicator disappeared mid-response on any request that took longer.

Add a TypingIndicator struct (mirroring the Discord implementation in PR #115) that re-sends the typing action every 8 s on a background thread. Wire it into the channel so the loop is started before processMessage and stopped via defer as soon as a reply is ready or an error occurs.

Ref: #111 (same issue for discord) which was fixed in #115

### DIFF
--- a/src/providers/claude_cli.zig
+++ b/src/providers/claude_cli.zig
@@ -8,7 +8,7 @@ const ChatMessage = root.ChatMessage;
 
 /// Provider that delegates to the `claude` CLI (Claude Code).
 ///
-/// Runs `claude -p <prompt> --output-format stream-json --model <model>`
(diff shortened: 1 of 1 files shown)


ibhagwan:
Also added https://github.com/nullclaw/nullclaw/pull/124/changes/12280b012820161b9cc6bf7e775d0cfb648c16b4, I think it’s important and also useful for the agent to know the context of the chat if its group chat or dm and the group_id, openclaw has that info and it also helps in configuration to know the exact group id so you can ask “what’s this group chat id” or let the agent configure itself (future PR).

Issue #136, opened 2026-02-27
Discord Gateway disconnect caused by tool loop ??


vaneistech:
### Description

I had several good conversations with the agent, and it was working fine. However, the bug appeared when I asked the agent to "prepare a script to install a program." 

Because the agent is in proactive mode, it attempted to execute the task using sudo and doas commands instead of making the script.

 The agent is working as a regular user (not root) and is in proactive mode maybe this caused the system to fail. 

The agent got stuck in a loop trying to run these commands until it exhausted the maximum tool iterations (25/25). Immediately after, the Discord connection failed with the following error:

`password: warning(discord): Discord gateway read failed: error.ConnectionClosed`
`warning(agent): Tool iterations exhausted (25/25), requesting summary`

### Expected behavior

not  to  drop connection 

### Steps to reproduce

Request the agent to prepare a script to install a program. 
The agent attempts to use sudo or doas commands. 
This time i am not sure it is my guess.

 You resolved the previous bugs super fast, and I really appreciate the effort.

### Version

56d519e

### OS

Linux

PR #138, opened 2026-02-27
fix(gateway): enforce security policy in daemon/channel runtime

DonPrus:


DonPrus:
Fixing https://github.com/nullclaw/nullclaw/issues/136

PR #143, opened 2026-02-27
signal: keep daemon default and add REST mode compatibility

TanGentleman:
## Summary
Makes Signal support for `signal-cli-rest-api` opt-in and non-breaking by keeping daemon JSON-RPC/SSE as the default path.

## Changes
- Preserves existing daemon behavior by default (`SIGNAL_USE_REST_API=0`).
- Adds REST send compatibility for `/v2/send` using `number`, `recipients`, and `base64_attachments` payload keys.
- Adds REST receive polling via `/v1/receive/{number}` when `SIGNAL_USE_REST_API=1`.
- Keeps health checks mode-aware (`/api/v1/check` for daemon, `/v1/health` for REST mode).
- Adds focused tests for REST URL builders and REST body schema fields.
- Fixes a WebSocket frame-unmasking compile issue in `signal.zig`.

## Why
This keeps current users on the stable daemon path while making it easy for developers to flip one env flag and run against dockerized `signal-cli-rest-api` in normal mode.

### DIFF
--- a/src/channels/signal.zig
+++ b/src/channels/signal.zig
@@ -1038,7 +1038,7 @@ pub const SignalChannel = struct {
 
         return try self.processEnvelope(
             allocator,
-            source orelse source_uuid,
+            source_uuid orelse source,
             source_number,
             source_name,
             timestamp_val,


DonPrus:
@TanGentleman  Is it ready for review?

TanGentleman:
@DonPrus I am considering adding it in the examples/ folder instead. I was also able to successfully use a cloudflare tunnel so the agent can communicate with a dockerized instance of the signal CLI, even when not sharing a network. Do you consider these use cases practical?

Nullclaw deployment is low-resource, so I'm hoping to be able to use CF tokens to connect to channels that are spun up elsewhere as needed for my use case.

I aimed to make this PR simpler, just throw Nullclaw and [Signal](https://github.com/bbernhard/signal-cli-rest-api/) into docker containers. I'll mark it ready for review, but it may need work!

PR #154, opened 2026-02-27
feat(config): add workspace_dir_override for user-specified workspace path

chenhlSH:


chenhlSH:
This change bundle adds proper support for specifying a custom workspace path in the interactive onboard wizard and ensures the chosen default model is persisted and honoured.

Key points:

workspace_dir_override field – new optional config entry.
Wizard writes it, Config.load() uses it, and save() emits valid JSON (now escaping backslashes).
Users can now point nullclaw at any directory (Windows paths included) instead of being stuck with ~/.nullclaw/workspace.

Model persistence bug fixed – the selected model string is duplicated before freeing the live list, preventing a dangling pointer.
Parsing of agents.defaults.model.primary already existed, but malformed JSON was being silently ignored; load now logs parse errors.

Improved config handling – Config.load() logs syntax errors, save() escapes values safely, and additional unit tests cover round‑trip save/load, workspace override parsing, and regression cases.

Test coverage – new tests verify the override field, proper round‑trip behaviour, and the original feature continues to work.

chenhlSH:
If it has already allowed users to choose where own workspace, then i supposed the fact that nullclaw still using default workspace is a kind of Lack of functionality. So i create a pr)

PR #166, opened 2026-02-28
fix(update): stream download to file to avoid memory limit

jd316:
## Summary
- Fix `nullclaw update` failing with `StdoutStreamTooLong` error when downloading large binary releases (~50MB+)
- Changed from `std.process.Child.run()` (buffers all stdout in memory) to streaming approach that writes directly to a temp file using a 64KB buffer
- Added proper cleanup of partial files on download failure

## Test plan
- [x] Run `nullclaw update` on a system with an older version installed
- [x] Verify download completes successfully for large binaries
- [x] Verify atomic replacement of binary works correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ reference/
 
 result
 .direnv/
+.worktrees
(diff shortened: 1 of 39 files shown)


jd316:
@DonPrus Hey! This PR is ready for review! 🙏

PR #255, opened 2026-03-03
Fix cron agent FileNotFound loop when self executable path is stale

manelsen:
## Summary
This PR adds a resilience fix for cron agent jobs that can fail with `FileNotFound` after in-place binary replacement on Linux.

## Problem
Cron agent jobs spawn a child process by re-executing the current binary path from `selfExePath`.
In some operational flows (for example rebuilding and restarting around the same path), the running process can expose a stale executable path that ends with ` (deleted)`.
When the scheduler later spawns agent jobs from that stale path, execution fails with `FileNotFound` in a loop.

## What changed
- `src/cron.zig`
  - Added `preferAgentExecPath()` to detect stale Linux self-exe paths and prefer `/proc/self/exe`.
  - In `runAgentJob()`:
    - retry once without `cwd` when spawn fails with `FileNotFound` (handles missing/invalid working dir);
    - on Linux, retry via `/proc/self/exe` when direct self path fails.
  - Added regression tests for executable-path selection behavior.

## Why this is safe
- Scope is isolated to cron agent subprocess spawn.
- Existing success path is unchanged when spawn works normally.
- Fallback logic only activates on `FileNotFound`.

## Validation
- `zig build`
  - Existing unrelated segfault in test runner remains in this environment (`session.test.concurrent processMessage different keys — no crash`), but compilation and test enumeration proceed.

### DIFF
--- a/src/cron.zig
+++ b/src/cron.zig
@@ -864,6 +864,10 @@ const AGENT_POLL_STEP_NS: u64 = 200 * std.time.ns_per_ms;
 const LINUX_SELF_EXE_PATH = "/proc/self/exe";
 const DELETED_EXE_SUFFIX = " (deleted)";
 
+fn pathAgentExecutableName() []const u8 {
+    return if (comptime builtin.os.tag == .windows) "nullclaw.exe" else "nullclaw";
+}
+
 fn hasTimeoutExpired(start_ns: i128, timeout_secs: u64) bool {
     if (timeout_secs == 0) return false;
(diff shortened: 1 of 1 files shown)


DonPrus:
There are no diff with `main`, did you `push -f` after I merge it?

manelsen:
Oops. Sorry. I might have pushed the cron fix along with the previous thing. Sorry.

PR #262, opened 2026-03-04
onboard: make AGENTS memory guidance backend-aware

### LINKED ISSUE #210: Better integration of AGENTS.md template and SQLite memories
### Description

Currently the AGENTS.md directs the agent to use Markdown files for storing and retrieving memories, even when SQLite is enabled.

### Motivation

Make SQLite memories more integrated.

manelsen:
## Summary
- update `src/workspace_templates/AGENTS.md` to be memory-backend-aware instead of markdown-only
- instruct agents to check `config.memory.backend` before loading/storing memory
- for sqlite/lucid/lancedb/postgres/redis/api/memory backends, guide agents to use memory tools (`memory_list`, `memory_recall`, `memory_store`)
- keep markdown file workflow for `markdown` backend and optional local notes
- add regression assertions in `src/onboard.zig` to ensure template keeps backend-aware memory guidance

## Why
Issue #210 reports that the AGENTS template always tells agents to use markdown memory files even when SQLite is enabled. That causes behavior drift from the configured memory backend.

This change aligns operational guidance with runtime memory backend selection.

## Validation
- `zig fmt src/onboard.zig`
- `zig build`
- `zig build test --summary all` (known unrelated existing crash in `session.test.concurrent processMessage different keys — no crash`; suite still reports `4787/4787 tests passed` before crash)

Fixes #210

### DIFF
--- a/src/agent/cli.zig
+++ b/src/agent/cli.zig
@@ -13,6 +13,7 @@ const tools_mod = @import("../tools/root.zig");
 const Tool = tools_mod.Tool;
 const memory_mod = @import("../memory/root.zig");
 const Memory = memory_mod.Memory;
+const bootstrap_mod = @import("../bootstrap/root.zig");
 const observability = @import("../observability.zig");
 const Observer = observability.Observer;
 const ObserverEvent = observability.ObserverEvent;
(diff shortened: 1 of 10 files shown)


DonPrus:
I tried to fix this, but then realized we also need to move AGENTS.md, HEARTBEAT.md, IDENTITY.md, SOUL.md, TOOLS.md, and USER.md into the DB as well. This definitely needs to be done, so tomorrow I will work on leveling up memory :)

manelsen:
We may move them inside the DB, or we may leave them out instead. I like the idea of having a hybrid default, to still have a functional, working, post-Hello-World assistant even if SQLite's file is corrupted. It's your call, of course.

PR #266, opened 2026-03-04
memory: preserve archive/conversation content before hygiene deletion

### LINKED ISSUE #263: Memory hygiene deletes files without summarization or embedding preservation
## Problem

The hygiene module (`src/memory/lifecycle/hygiene.zig`) permanently deletes memory files and conversation rows without preserving their semantic content first. The summarizer module exists (`src/memory/lifecycle/summarizer.zig`) but is completely disconnected from the hygiene pipeline.

### Current behavior

1. `archiveOldFiles()` moves `memory/*.md` to `memory/archive/` after 7 days (default)
2. `purgeOldArchives()` **deletes** archived files after 30 days — permanently gone
3. `pruneConversationRows()` calls `mem.forget()` on conversation entries older than 30 days — permanently gone

None of these steps embed, summarize, or extract key facts before deletion. The summarizer module is only used during active conversations for context window management — not for long-term preservation.

### Why this matters

- **Memory loss is the default.** An agent running for 2+ months silently loses all memories older than 30 days.
- **Vector storage is cheap.** 2,000 vectors at 2048 dimensions ≈ 16 MB. A year of daily memories would fit in <100 MB of vector storage — negligible even on a $5 ARM board.
- **The summarizer already exists** but is not wired into the lifecycle. The `hygiene.zig` module has zero imports from `summarizer.zig`, `embeddings.zig`, or any vector store module.
- **The episodic/semantic split in the summarizer is exactly what's needed** — extract "Key fact:" entries as permanent core memories before archiving the verbose source.

### Suggested fix

Add a pre-deletion preservation step to the hygiene pipeline:

```
Before purge:
  1. Read the file content
  2. Chunk and embed into the vector store (if not already indexed)
  3. Optionally: run summarizer to extract key facts → store as core memories
  4. Then delete the source file
```

#### Minimal implementation

In `purgeOldArchives()`, before `deleteFile`:

```zig
// Preserve semantic content before purging
if (vector_store) |vs| {
    const content = archive_dir.readFileAlloc(allocator, entry.name, 1024 * 1024) catch continue;
    defer allocator.free(content);
    const chunks = chunker.chunkMarkdown(allocator, content, .{}) catch continue;
    defer chunker.freeChunks(allocator, chunks);
    for (chunks) |chunk| {
        vs.upsert(chunk.text, .{ .source = entry.name, .archived = true }) catch {};
    }
}
```

#### Full implementation (with summarization)

```zig
// Before purge, summarize and extract permanent facts
if (summarizer_config.enabled) {
    const content = readFile(entry.name);
    const prompt = summarizer.buildSummarizationPrompt(allocator, content);
    // caller handles LLM call, then:
    const result = summarizer.parseSummaryResponse(allocator, llm_response);
    for (result.extracted_facts) |fact| {
        if (fact.category == .core) {
            mem.store(fact.key, fact.content, .core, null);
        }
    }
}
```

### Config suggestion

```json
{
  "memory": {
    "hygiene_enabled": true,
    "archive_after_days": 7,
    "purge_after_days": 30,
    "preserve_before_purge": true,
    "summarize_before_purge": false
  }
}
```

- `preserve_before_purge: true` — embed chunks into vector store before deleting source files (cheap, no LLM needed)
- `summarize_before_purge: false` — optionally run LLM summarization for denser representations (costs API calls)

### Related modules

- `src/memory/lifecycle/summarizer.zig` — has the right abstractions, just needs to be called
- `src/memory/vector/chunker.zig` — already handles markdown chunking
- `src/memory/vector/store.zig` — VectorStore interface ready for upserts
- `src/memory/vector/embeddings.zig` — EmbeddingProvider vtable

The building blocks are all there. They just need to be connected.

manelsen:
## Summary
- add `memory.lifecycle.preserve_before_purge` (default: `true`) to lifecycle config and parser
- preserve archived markdown content before purge by chunking and storing it into memory under `archive:*` keys (`category=archive`)
- preserve old conversation rows before prune by chunking/storing under `archive:conversation:*` keys, then forgetting the original row
- fail safe: if preservation fails, hygiene skips deletion/prune for that item to avoid silent data loss
- add an optional hygiene preserve hook and wire it into archive + conversation preservation paths
- execute hygiene after vector plane setup so preserved chunks can be indexed immediately
- sync preserved chunks into vector search using the existing strategy:
  - durable mode: enqueue `upsert` in outbox
  - best-effort mode: embed + `vector_store.upsert`
- add integration test coverage that proves preserved archive chunks enqueue durable outbox jobs during `initRuntime`

## Why
Issue #263 reports that hygiene permanently deletes archived files and conversation rows without preserving semantic content first. The first patch added preservation before deletion. This follow-up ensures those preserved chunks are also available to semantic/vector retrieval without waiting for later backfill.

## Config
```json
{
  "memory": {
    "lifecycle": {
      "preserve_before_purge": true
    }
  }
}
```

## Validation
- `zig build` ✅
- `zig build test --summary all` ⚠️ existing unrelated failure remains: `session.test.concurrent processMessage different keys — no crash` (signal 11)

Closes #263

### DIFF
--- a/src/memory/lifecycle/hygiene.zig
+++ b/src/memory/lifecycle/hygiene.zig
@@ -268,7 +268,7 @@ fn preserveConversationEntry(
 }
 
 /// Prune conversation rows older than retention_days via the Memory interface.
-/// Searches for conversation-tagged entries and deletes those whose timestamp is old.
+/// Lists conversation-tagged entries and deletes those whose timestamp is old.
 pub fn pruneConversationRows(allocator: std.mem.Allocator, mem: Memory, retention_days: u32) !u64 {
     return pruneConversationRowsWithPreserve(allocator, mem, retention_days, false, null);
 }
@@ -282,8 +282,9 @@ fn pruneConversationRowsWithPreserve(
 ) !u64 {
     const cutoff_secs = std.time.timestamp() - @as(i64, @intCast(retention_days)) * 24 * 60 * 60;
 
-    // Search for conversation-tagged entries
-    const results = mem.search(allocator, "conversation", 1000) catch return 0;
+    // List conversation-tagged entries directly so prune is independent of
+    // message text content.
+    const results = mem.list(allocator, .conversation, null) catch return 0;
     defer {
         for (results) |r| r.deinit(allocator);
         allocator.free(results);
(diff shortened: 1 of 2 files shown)


manelsen:
Should I close the PR? Will you squash and merge? What would you like me to do?

PR #271, opened 2026-03-04
add modal-slack example

TanGentleman:
## Summary
- Adds a complete Modal-based Slack example for multi-agent deployment (`examples/modal-slack/`)
- Includes setup script (`slack_setup.py`) that creates a private channel, invites bots, and generates config/env files
- Includes `modal_app.py`, deploy/install scripts, config templates, and README

## Details
Mirrors the existing `modal-matrix` example but for Slack using Socket Mode. The setup script handles:
- Slack API auth verification for planner and builder bots
- Private channel creation with fallback to lookup if channel already exists
- `--channel-id` flag for using an existing channel
- `--dry-run` mode for previewing config without writing files

## Test plan
- [ ] Run `slack_setup.py --dry-run` with valid bot tokens to verify config output
- [ ] Run full setup to create channel and generate `config.slack.json` / `.env`
- [ ] Deploy with `modal deploy` and verify both bots connect via Socket Mode

### DIFF
--- a/examples/modal-slack/config.slack.example.json
+++ b/examples/modal-slack/config.slack.example.json
@@ -57,7 +57,7 @@
     }
   },
 
-  "agent_bindings": [
+  "bindings": [
     { "agent_id": "planner", "match": { "channel": "slack", "account_id": "planner-account" } },
     { "agent_id": "builder", "match": { "channel": "slack", "account_id": "builder-account" } }
   ],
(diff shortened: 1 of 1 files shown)


DonPrus:
I'm currently implementing orchestration in https://github.com/nullclaw/NullBoiler to keep the client simple and avoid adding extra complexity here. I can rework this PR a bit later in that repository.

DonPrus:
Something like this: https://github.com/nullclaw/nullboiler/tree/main/examples/multi-agent-slack

PR #276, opened 2026-03-04
fix(openai-codex): send request body via stdin to curl

gula00:
## Summary

- Fix openai-codex request delivery to curl by sending the JSON payload through stdin instead of command-line arguments.
- Replace -d <body> with --data-binary @- and write the request body via child.stdin.
- Keep provider behavior unchanged aside from transport robustness.

## Why
openai-codex requests can produce large JSON payloads.  

Passing the full body via argv (curl -d ...) can exceed OS command-line limits, especially on Windows.
- Windows CreateProcessW command-line limit: 32,767 characters (documented by Microsoft: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw)
- macOS ARG_MAX (typical): 262,144 bytes

In repro, request bodies reached ~32,223 and ~44,265 bytes; with additional curl args/headers, this can exceed Windows limits and surface as AllProvidersFailed.

Using stdin avoids argv size limits and makes request submission robust across platforms.

## Changes
- Updated src/providers/openai_codex.zig in codexStreamRequest(...):
  - switch curl payload mode to --data-binary @-
  - enable piped stdin for the child process
  - write the body to stdin and close it before reading SSE output

## Validation
- zig build -Doptimize=Debug
- Reproduced successful run after fix:
  - nullclaw agent --provider openai-codex --model gpt-5.3-codex -m "Hello!"

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 The smallest fully autonomous AI assistant infrastructure — a static Zig binary that fits on any $5 board, boots in milliseconds, and requires nothing but libc.
 
 ```
-678 KB binary · <2 ms startup · 3,230+ tests · 22+ providers · 18 channels · Pluggable everything
+678 KB binary · <2 ms startup · 3,230+ tests · 23+ providers · 18 channels · Pluggable everything
 ```
 
 ### Features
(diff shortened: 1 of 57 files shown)


gula00:
### Follow-up update

This PR now also includes the same argv-limit hardening for additional provider paths that still passed large JSON bodies via `curl -d`:
- `src/providers/gemini.zig` (Gemini streaming path)
- `src/providers/anthropic.zig` (`curlPostOAuth`, used by Claude Code OAuth flow)

PR #287, opened 2026-03-04
Fix: URL Special Characters (`?` and `&`) in Command Strings

brunojuliao:
It's me again 😅 My nullclaw had issues using '?' and '&' in the URL. I've granted full autonomy. After investigating with Claude Code, we came up to this enhacement, which is opt-in (feel free to edit or request changes as you see fit):

# Fix: URL Special Characters (`?` and `&`) in Command Strings

## Problem

Even with `"allowed_commands": ["*"]`, nullclaw's security policy rejects commands
containing `?` and `&` in the command string. This breaks legitimate use cases like:

```bash
curl "https://api.example.com/search?q=test&page=1&limit=10"
```

The workaround was to use the `-G` flag (e.g. `curl -G url -d key=val`), which
avoids placing `?` and `&` directly in the raw command string. This workaround
should not be necessary for quoted URLs.

## Root Cause

The `containsSingleAmpersand()` function in `src/security/policy.zig` scans the
**entire raw command string** for standalone `&` characters (i.e. `&` that is not
part of `&&`). It was not quote-aware: a `&` inside double or single quotes (like
a URL query string) was treated identically to a `&` shell background operator.

```
curl "https://example.com?a=1&b=2"
                               ^
                               This & is inside quotes and is NOT a shell
                               operator, but was incorrectly flagged.
```

The `?` character itself was **not blocked** for general commands (only for
`rm`/`trash` bootstrap-delete safety checks). However, URLs with query strings
typically contain both `?` and `&`, so in practice both characters appeared to
be rejected.

### 1. Quote-aware `containsSingleAmpersand` (default behavior change)

The function now tracks single-quote and double-quote state while scanning. Any
`&` character found inside a quoted region is skipped. This means:

| Command | Before | After |
|---------|--------|-------|
| `curl "https://example.com?a=1&b=2"` | Blocked | Allowed |
| `curl 'https://example.com?a=1&b=2'` | Blocked | Allowed |
| `curl https://example.com?a=1&b=2` | Blocked | Blocked (unquoted `&`) |
| `ls & rm -rf /` | Blocked | Blocked (shell operator) |
| `ls && echo done` | Allowed | Allowed (`&&` is safe) |

This is a **correctness fix**: `&` inside quotes is never a shell background
operator, so the old behavior was a false positive.

### 2. `allow_raw_url_chars` config option (opt-in)

For users who need bare (unquoted) `&` in commands (e.g. when the shell has
already stripped quotes), a new config option completely bypasses the `&` check:

```json
{
  "autonomy": {
    "allow_raw_url_chars": true
  }
}
```

- **Default: `false`** (secure by default; unquoted `&` is still blocked)
- When `true`: skips the `containsSingleAmpersand` check entirely
- All other safety checks remain enforced (subshell injection, redirects,
  process substitution, tee blocking, high-risk command gating, etc.)

### 3. What is NOT changed

- `?` handling in `isSafeBootstrapDeleteTarget` remains strict (glob prevention
  for `rm`/`trash` commands targeting `BOOTSTRAP.md`)
- `&&` (double ampersand) was already allowed and remains so
- All other security checks (`$(...)`, backticks, `>` redirects, `<(...)`,
  `>(...)`, `tee`, high-risk commands, rate limiting) are untouched

## Files Changed

| File | Change |
|------|--------|
| `src/security/policy.zig` | Quote-aware `containsSingleAmpersand`, `allow_raw_url_chars` field, new tests |
| `src/config_types.zig` | `allow_raw_url_chars` field in `AutonomyConfig` |
| `src/config_parse.zig` | JSON parsing for `allow_raw_url_chars` |
| `src/gateway.zig` | Pass `allow_raw_url_chars` to `SecurityPolicy` |
| `src/channel_loop.zig` | Pass `allow_raw_url_chars` to `SecurityPolicy` |
| `config.example.json` | Document the new option |

## Test Coverage

New tests added to `src/security/policy.zig`:

- `containsSingleAmpersand_skips_quoted_ampersands` - unit test for quote-aware detection
- `quoted_url_with_ampersand_passes_wildcard_allowlist` - end-to-end: quoted URLs pass with `["*"]`
- `unquoted_url_with_ampersand_blocked_by_default` - bare `&` still blocked without opt-in
- `allow_raw_url_chars_permits_bare_ampersand` - opt-in flag allows bare `&`
- `allow_raw_url_chars_still_enforces_other_safety_checks` - other guards remain active
- `allow_raw_url_chars_defaults_to_false` - secure default
- `dash_G_workaround_avoids_special_chars` - documents the `-G` workaround still works

### DIFF
--- a/src/providers/compatible.zig
+++ b/src/providers/compatible.zig
@@ -346,6 +346,68 @@ pub const OpenAiCompatibleProvider = struct {
         needs_free: bool,
     };
 
+    const SplitThinkContent = struct {
+        visible: []const u8,
+        reasoning: ?[]const u8,
+    };
+
+    fn scanThinkContent(
+        allocator: std.mem.Allocator,
+        text: []const u8,
+        visible_out: *std.ArrayListUnmanaged(u8),
+        reasoning_out: ?*std.ArrayListUnmanaged(u8),
+    ) !void {
+        var i: usize = 0;
+        var depth: usize = 0;
+        while (i < text.len) {
+            if (std.mem.startsWith(u8, text[i..], think_open_tag)) {
+                depth += 1;
+                i += think_open_tag.len;
+                continue;
+            }
+            if (std.mem.startsWith(u8, text[i..], think_close_tag)) {
+                if (depth > 0) depth -= 1;
+                i += think_close_tag.len;
+                continue;
+            }
+
+            if (depth == 0) {
+                try visible_out.append(allocator, text[i]);
+            } else if (reasoning_out) |reasoning_buf| {
+                try reasoning_buf.append(allocator, text[i]);
+            }
+            i += 1;
+        }
+    }
+
+    fn splitThinkContent(allocator: std.mem.Allocator, text: []const u8) !SplitThinkContent {
+        if (std.mem.indexOf(u8, text, think_open_tag) == null and std.mem.indexOf(u8, text, think_close_tag) == null) {
+            return .{
+                .visible = try allocator.dupe(u8, text),
+                .reasoning = null,
+            };
+        }
+
+        var visible_buf: std.ArrayListUnmanaged(u8) = .empty;
+        defer visible_buf.deinit(allocator);
+        var reasoning_buf: std.ArrayListUnmanaged(u8) = .empty;
+        defer reasoning_buf.deinit(allocator);
+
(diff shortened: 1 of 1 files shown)


brunojuliao:
Sorry for missing the last merge. I'm in a class 😅 Thank you for merging 😇

PR #299, opened 2026-03-05
feat(telegram): stream partial responses via sendMessageDraft

juvenn:
### Summary

Telegram Bot API supported partial streaming message to **all bot** since [March 1st](https://core.telegram.org/bots/api#march-1-2026).

Leverage Telegram's new [sendMessageDraft Bot API](https://core.telegram.org/bots/api#sendmessagedraft) to show users a live-updating draft while the AI response is being generated, instead of only displaying a typing indicator.

### Change Overview

- Add DraftState, sendDraft, and vtableSendEvent to TelegramChannel with per-chat draft buffers flushed at 64-byte / 200ms thresholds
- Wire streaming sink into channel_loop's processTelegramMessage via makeSink/StreamCtx so the agent loop streams through the channel
- Delegate supports_streaming and stream_chat in ReliableProvider to the inner provider so the agent detects streaming capability
- Enable streaming outbound path for telegram in daemon (webhook mode)
- Add streaming config toggle (default: true) to TelegramConfig

### Config example

```json
{
  "channels": {
    "telegram": {
      "enabled": true,
      "streaming": false, // default true
     ...
    }
  }
}
```

### Discussions

~~One remaining issue: streaming mode is causing raw <tool_calls> to render in the chat. Would you prefer I address this before the merge, or should we track it as a follow-up?~~

EDIT, issue addressed. Added a stream tag filter to sanitize tool_call tags. ✅

### DIFF
--- a/src/channel_loop.zig
+++ b/src/channel_loop.zig
@@ -88,25 +88,30 @@ fn processTelegramMessage(
             error.OutOfMemory => "Out of memory.",
             else => "An error occurred. Try again or /new for a fresh session.",
         };
+        if (sink != null) {
+            tg_ptr.channel().sendEvent(sender, "", &.{}, .final) catch {};
+        }
         tg_ptr.sendMessageWithReply(sender, err_msg, reply_to_id) catch |send_err| log.err("failed to send error reply: {}", .{send_err});
         return;
     };
     defer allocator.free(reply);
 
     if (shouldSuppressGroupReply(is_group, reply)) {
+        if (sink != null) {
+            tg_ptr.channel().sendEvent(sender, "", &.{}, .final) catch {};
+        }
         log.info("Smart reply: skipping non-essential message", .{});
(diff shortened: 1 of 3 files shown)


nordlauf:
Hi, I am a bit confused about the config example above. My config structure looks differently. Where exactly should `streaming: false` go?

My config structure

```
    "channels": {
        "telegram": {
            "accounts": {
                "default": {
                    "account_id": "default",
                    "bot_token": "8888redacted-redacted",
                    "allow_from": ["7XXXXXXXX"],
                    "group_allow_from": [],
                    "group_policy": "allowlist",
                    "reply_in_private": true,
                    "proxy": null,
                    "interactive": {
                        "enabled": false,
                        "ttl_secs": 900,
                        "owner_only": true,
                        "remove_on_click": true
                    }
                }
            }
        }
    },

```

PR #333, opened 2026-03-06
fix(telegram): debounce split long messages across getUpdates polls

### LINKED ISSUE #312: Telegram: need debounce window (2–5s) to coalesce split long messages across getUpdates polls
### Description

When sending long messages from Telegram, the message is sometimes split by Telegram into multiple consecutive messages (Telegram limit ~4096 chars).

Even though the repository includes the message merging logic from PR #145 (mergeConsecutiveMessages), NullClaw still replies to each chunk individually instead of waiting and responding once with the full merged message.

1 user message → split into 3 Telegram messages
→ agent produces more than one separate replies (sometimes 3, sometimes 2, but more than 1 all the time, at least in my experience).

From inspecting the runtime behavior, it appears that Telegram updates are processed as soon as they arrive via getUpdates polling. Because getUpdates returns as soon as the first message arrives, the chunks often land in separate polls, meaning they never appear together in the same batch where mergeConsecutiveMessages could merge them.

For context, NullClaw is running as a service.

Telegram polling is visible because the process is calling the Bot API getUpdates.

So even though the merge logic exists, in real usage long Telegram messages still produce multiple agent responses.

### Expected behavior

If multiple text messages arrive from the same user and chat within a short time window, the system should wait briefly before processing and merge them into a single logical message.

A small configurable debounce window (e.g. 2–5 seconds per chat/session) before processing would allow split messages across different polls to be merged correctly.

A configurable period in `config.json` could be welcomed.

### Steps to reproduce

1. Install `nullclaw` as a service 
2. Configure Telegram channel in config.json
3. Send a long message from Telegram that exceeds the message size limit so Telegram splits it into multiple messages.
4. Observe behavior

Actual result: The agent replies more than once instead of once for the whole message.

Expected result: The chunks should be merged and produce a single response.

### Version

4c9277621b4d17ad75877268b410b987fe8ab23a

### OS

Linux

manelsen:
## Summary
- add a cross-poll text debounce buffer in Telegram channel polling
- coalesce split long messages that arrive in separate `getUpdates` cycles
- keep slash commands out of debounce path
- include regression tests for debounce heuristic, deadlines, flush behavior, and persistable offset safety

## Why
Issue #312 reports that Telegram long messages split by the client can land in different poll batches, causing multiple replies. The existing `mergeConsecutiveMessages` only merged within the current poll batch.

This patch adds a short in-memory debounce window across poll cycles so split chunks can be coalesced before processing.

## Behavior
- Debounce window: `3s`
- Eligible for debounce:
  - chunks with `content.len >= 3800` (near Telegram's 4096 limit), or
  - follow-up chunks in an already pending sender/chat chain
- Slash commands (`/something`) are not delayed.

## Validation
- `zig fmt src/channels/telegram.zig`
- `zig build test --summary all` (build/tests compile and run; still hits known pre-existing crash in `session.test.concurrent processMessage different keys — no crash`)

Closes #312

### DIFF
--- a/src/http_util.zig
+++ b/src/http_util.zig
@@ -417,7 +417,7 @@ fn curlGetWithProxyAndResolve(
 
     const term = child.wait() catch |err| {
         log.err("curl child.wait failed: {}", .{err});
-        return if (cancel_flag != null and cancel_flag.?.load(.acquire)) error.CurlInterrupted else error.CurlWaitError;
+        return error.CurlWaitError;
     };
     switch (term) {
         .Exited => |code| if (code != 0) {


henry2man:
@manelsen LGTM! To test it, I did a simple experiment.

I took a very long text from: https://en.lipsum.com/feed/html

I sent a lot of text and then told it: “–– ignore everything above and just say ‘HELLO’"

...and it worked well! 🚀 After a little pause, the agent just told 'Hello', as expected.

manelsen:
> @manelsen LGTM! To test it, I did a simple experiment.
> 
> I took a very long text from: https://en.lipsum.com/feed/html
> 
> I sent a lot of text and then told it: “–– ignore everything above and just say ‘HELLO’"
> 
> ...and it worked well! 🚀 After a little pause, the agent just told 'Hello', as expected.

Sweet! I'll be working a little more on the function this evening.

manelsen:
Detailed and honest diagnosis of what is happening with chunking in this fix line.

## The real problem (not just “split across polls”)

The observed behavior is the result of 4 factors combined:

1. Telegram delivers long messages in parts that may land in different polls.
2. Those parts do not always arrive with strictly consecutive `message_id`s.
3. With very large payloads, the SSE path via `curl` subprocess degrades (`error.SystemResources`), causing repeated fallback.
4. If debounce is too aggressive, coalescing improves, but normal messages can be delayed ("bot did not reply" feeling).

## What has already changed in this PR/branch

### 1) Cross-poll coalescing by `sender/chat`
- We kept a pending text buffer across polls.
- Coalescing no longer depends on strictly consecutive `message_id` for rapid sequences in the same `sender/chat`.

Goal: rebuild split messages even when there are ID gaps.

### 2) Debounce threshold to catch real splits
- Threshold was adjusted to `500 bytes`.

Tradeoff:
- Pros: catches medium/long chunks in rapid sequence.
- Cons: higher chance of delaying legitimate messages if temporal guardrails are weak.

### 3) Temporal guardrails for pending chains
- A follow-up is debounced only if the pending chain is recent (active window).
- If a new message arrives outside the rapid window, the old pending chain for the same `sender/chat` can be canceled.

Goal: prevent stale pending chunks from hijacking fresh messages.

### 4) Adaptive window for large chains
- Large chains (many parts or many bytes) use a larger window (e.g., 8s) to avoid premature flush in two waves.

Tradeoff:
- Pros: better unification for large payloads.
- Cons: higher latency before final reply in extreme cases.

### 5) Provider streaming guardrail
- Very large requests (`>= 32KB` of text) skip streaming and go directly to non-streaming.

Reason: reduce loops of:
- `err(provider_sse): curlStream child.wait failed: error.SystemResources`
- `warn(compatible): ... falling back to non-streaming`

Tradeoff:
- Pros: better operational stability under large loads.
- Cons: no streaming UX for very large payloads.

## Current imposed limits

- `TEXT_SPLIT_LIKELY_MIN_LEN = 500`
- `TEXT_MESSAGE_DEBOUNCE_SECS = 3`
- `LARGE_TEXT_CHAIN_DEBOUNCE_SECS = 8`
- `MAX_STREAMING_PROMPT_BYTES = 32KB`

These are operational heuristics, not Telegram semantic guarantees.

## Explicit architectural limitation

Current architecture prioritizes broad compatibility (OpenAI-compatible + fallbacks), but at a cost:

- Critical streaming still depends on `curl` subprocesses.
- There is no transactional chunk assembler with explicit completeness acknowledgment.
- Debounce is inherently heuristic; there is no perfect "last chunk" signal from the channel.

For extreme payloads (e.g., ~99k), the system is balancing:
- fast replies (risk: fragmentation), and
- waiting longer for coalescing (risk: perceived latency).

## Recommended next steps

1. Make limits configurable in `config`:
- base debounce
- large-chain debounce
- oversized streaming cutoff

2. Separate channel policy and provider policy clearly (Telegram chunking vs provider streaming).

3. Move critical streaming to a native HTTP path (reduce `curl` subprocess dependence under high load).

Issue #314, opened 2026-03-05
Telegram: "Tokens used" always shows 0


nordlauf:
### Description

<img width="720" height="537" alt="Image" src="https://github.com/user-attachments/assets/69c1c40b-8c21-4d46-a341-2ff55e8a3bee" />

### Expected behavior

It should show tokens used.

### Steps to reproduce

Do "/status" and check Tokens used.

### Version

2026.3.4

### OS

Linux

PR #340, opened 2026-03-06
fix(status,threads): estimate fallback tokens and align runtime stack budgets

manelsen:
## Summary
- fix Issue #314 where `/status` could keep showing `Tokens used: 0` when provider usage fields are absent
- add fallback token estimation in agent turn flow when `prompt_tokens`, `completion_tokens`, and `total_tokens` are all zero
- align threaded session-turn stack sizing with the runtime budget used by channel workers
- centralize thread stack budgets in `src/thread_stacks.zig` and replace repeated literals by runtime role
- keep runtime and concurrency tests on the same named session-turn stack budget to reduce future drift

## Why
- some providers and channels do not return usage counters, so `/status` could remain at zero indefinitely without fallback estimation
- the concurrent session tests were using a smaller stack than the runtime worker path, which made the failure mode look like a session crash instead of a test/runtime stack-budget mismatch

## Validation
- `zig build`

## Notes
- this PR now also includes the thread-stack follow-up merged in `manelsen/nullclaw#1`


manelsen:
🌊 NullClaw 2026.3.5
🧠 Model: glm-4.7
💬 History: 35 messages
🧮 Tokens used: 215939
🔧 Tools: 18 available
💭 Thinking: off
📢 Verbose: off
🧩 Reasoning: off
📈 Usage: off
⚙️ Exec: host=gateway security=full ask=off
🪢 Queue: mode=off debounce=0ms cap=0 drop=summarize
🔊 TTS: mode=off provider=default
📡 Activation: mention
📤 Send: inherit
⏰ Session TTL: off

PR #361, opened 2026-03-08
Add daemon auto model routing support

### LINKED ISSUE #360: feat: auto-router for models using task complexity, capabilities, and degraded-provider state
### Description

I can already configure:

- `agents.list`
- `reliability.fallback_providers`
- `reliability.model_fallbacks`
- `model_routes`

But I still have to choose the initial route manually.

Right now nullclaw has decent post-selection failover, but it still lacks real pre-selection auto-routing for the initial model/provider choice.

### Current behavior

Today the choices are basically:

- set one default model
- explicitly choose a named agent profile
- rely on fallback only after the first route fails

That is not enough when I want:

- cheap/fast models for simple work
- stronger models for hard debugging/refactors/architecture tasks
- vision-capable routes for image work
- automatic skipping of temporarily degraded providers before the user sees a failed turn

Also, `model_routes` already exists in config, but it does not appear to behave like a real runtime auto-router yet.

### Expected behavior

Nullclaw should be able to choose the initial route automatically based on the current turn.

Examples:

- short/plain questions -> `fast`
- normal coding/config/build work -> `balanced`
- deep debugging, regressions, refactors, architecture -> `deep`
- image/screenshot/vision tasks -> `vision`
- recently rate-limited / quota-limited provider-model route -> skip temporarily and choose the next valid route

### Proposed scope

I am not asking for a full rule DSL in the MVP.

Smallest useful implementation:

- add a small `router` config section
- make `model_routes` an actual initial-route map
- allow `model_routes[].agent` so routes can resolve through `agents.list`
- apply it to main agent turns
- apply it to subagent spawning when no explicit agent is provided
- keep `ReliableProvider` responsible for retries / provider fallback / model fallback after the initial route is chosen

### Example config shape

```json
{
  "router": {
    "enabled": true,
    "default_hint": "balanced",
    "degrade_ttl_secs": 900,
    "long_context_tokens": 120000,
    "deep_turn_threshold": 12,
    "prefer_named_agents": true
  },
  "model_routes": [
    { "hint": "fast", "agent": "fast" },
    { "hint": "balanced", "agent": "balanced" },
    { "hint": "deep", "agent": "deep" },
    { "hint": "vision", "provider": "openrouter", "model": "google/gemini-2.5-pro" }
  ]
}
```

### Why this matters

If you use multiple providers and multiple models, the current setup either:

- wastes expensive/slow models on simple work
- requires manual switching too often
- or waits until a route fails before trying something more appropriate

Native routing would make nullclaw much more usable for mixed fast/balanced/deep setups, especially when providers have rate limits, credits, or capability differences.

### Additional context

I already prototyped this locally with a wrapper around `nullclaw agent`, and it works well enough as a stopgap. But this should live in nullclaw itself so it works consistently for:

- main agent turns
- subagents
- channels
- future gateway/web usage

juslintek:
Closes #360

## Summary
- add agent-side automatic model selection from `model_routes` based on prompt, context length, and image turns
- pin routing off for explicit `/model`, `/reset <model>`, `/restart <model>`, and CLI `--model` / `--provider` overrides
- add `/model auto` to re-enable automatic routing for the current session
- route runtime calls through concrete `provider/model` refs in the daemon path
- extend provider routing so tool calls, streaming, warmup, and explicit provider refs work consistently
- document `model_routes` and interactive routing behavior in `docs/en/` and `docs/zh/`

## Validation
- `git diff --check`

## Notes
- adds regression coverage for route selection, pinning behavior, `/model auto`, routed model propagation, and router-backed runtime behavior
- adds a micro-benchmark guard to keep the route selection path below visible per-turn overhead
- keeps routing in the daemon/runtime path instead of relying on an external wrapper

### DIFF
--- a/docs/en/commands.md
+++ b/docs/en/commands.md
@@ -49,7 +49,8 @@ This page groups the NullClaw CLI by task so you can find the right command quic
 
 - In `nullclaw agent`, `/model` shows the current model plus configured routing/fallback status.
 - `/model <provider/model>` pins the current session to that model and disables automatic routing.
-- `/model auto` clears the user pin and re-enables `model_routes` for later turns in the same session.
+- `/model auto` clears the user pin, restores the configured default model, and re-enables `model_routes` for later turns in the same session.
+- If no `model_routes` are configured, `/model auto` still clears the pin and returns the session to the configured default model.
 - Starting `nullclaw agent` with `--model` or `--provider` also pins the run and bypasses `model_routes`.
 
 ## Runtime and operations
--- a/docs/zh/commands.md
+++ b/docs/zh/commands.md
@@ -35,7 +35,8 @@
 
 - 在 `nullclaw agent` 里，`/model` 会显示当前模型以及已配置的路由/回退状态。
 - `/model <provider/model>` 会把当前会话 pin 到该模型，并关闭自动路由。
-- `/model auto` 会清除这个用户 pin，让同一会话后续回合重新使用 `model_routes`。
+- `/model auto` 会清除这个用户 pin，把会话恢复到配置里的默认模型，并让后续回合重新使用 `model_routes`。
+- 如果没有配置 `model_routes`，`/model auto` 仍然会清除 pin，并把会话切回配置里的默认模型。
 - 通过 `--model` 或 `--provider` 启动 `nullclaw agent` 时，也会把该次运行 pin 到显式模型，从而绕过 `model_routes`。
 
 ## 运行与运维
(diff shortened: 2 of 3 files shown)


juslintek:
Closing this implementation PR and keeping issue #360 as the public feature request. I am keeping the implementation private/local for now.

juslintek:
Superseded by #405. GitHub would not reopen this PR after the head branch was deleted and recreated, so the auto-routing change was restored as a single-commit replacement PR.

Issue #356, opened 2026-03-07
Windows shell tool output with non-UTF-8 / GBK UnicodeEncodeError can poison agent session and lead to subsequent error.CurlFailed


yuanmao1:
### Description

On Windows, invoking a subprocess through the `shell` tool that prints Unicode characters (such as emoji) to a GBK console can cause the subprocess to fail with a Python `UnicodeEncodeError`.

In my case, the agent was asked to run a Python script (`nullclaw/pet.py`) which prints emoji such as `🐾` and `❌`. Because the Windows console was using the default GBK code page, Python raised:

```
UnicodeEncodeError: 'gbk' codec can't encode character ...
```

This initial failure is expected for the script itself, but after the tool call fails, the current agent session becomes unstable.

From that point onward, **any subsequent user input causes the agent request to fail with**:

```
error.CurlFailed
```

Based on the debug logs, it appears that the shell tool output may contain bytes that are not normalized to valid UTF-8 before being appended to the conversation history. As a result, malformed content may be inserted into the message history, which later causes request serialization or transport to fail.

In other words, a subprocess encoding failure seems able to **poison the current session**, causing unrelated future prompts to fail.

Relevant log snippet:

```
UnicodeEncodeError: 'gbk' codec can't encode character '\U0001f43e'
```

After that:

```
Error: error.CurlFailed
```

Additionally, the logged message content appears as a byte array instead of a normal string:

```
content=[91,84,111,111,108,...]
```

which suggests the tool result may not have been safely decoded before being appended to the message list.

### Expected behavior

A subprocess encoding failure should remain a **normal tool error** and should not corrupt the agent session.

After a failed `shell` tool invocation, the agent should still be able to process later prompts normally.

Tool outputs should always be normalized into valid text before being inserted into the conversation history.

### Steps to reproduce

Environment:

* Windows
* default console code page (GBK)
* nullclaw running locally

Steps:

1. Create a Python script that prints emoji (for example `🐾` or `❌`).
2. Ask `nullclaw agent` to execute the script via the shell tool.
3. Allow the script to fail with a `UnicodeEncodeError` caused by GBK console encoding.
4. After the tool call fails, send another unrelated prompt to the agent.

Result:

The agent fails with:

```
error.CurlFailed
```

and continues failing for subsequent prompts within the same session.

Restarting the session clears the problem.

### Version

v2026.3.7

### OS

Other

PR #367, opened 2026-03-08
Fix tool call response ASCII codes and Windows GBK encoding issues

qxo:
1. Fixed tool call response content showing ASCII codes instead of strings by removing std.json.fmt from log statements
2. Fixed Windows GBK encoding garbled text issue in shell tool by:
   - Forcing UTF-8 codepage with 'chcp 65001' for cmd.exe
   - Setting Python UTF-8 environment variables (PYTHONUTF8, PYTHONIOENCODING)

This ensures proper UTF-8 output for shell commands on Windows Chinese systems.


yuanmao1:
This issue has already been raised in issue #356  and fixed in PR #357

qxo:
ok

PR #382, opened 2026-03-09
fix(discord): accept replies to bot when require_mention is enabled

### LINKED ISSUE #337: Discord: reply-to-bot messages ignored when require_mention is true
## Summary

When `require_mention: true` is set in Discord channel config, replying to the bot's own message (using Discord's Reply feature) does not trigger a response. Only explicit `@mention` in message text works.

## Expected Behavior

When a user replies to a message authored by the bot, the bot should treat it as a directed interaction and respond — even with `require_mention: true`. This matches the behavior of other Discord bots (easyclaw, MEE6, etc.) and is the natural UX expectation.

## Current Behavior

`isMentioned()` in `src/channels/discord.zig` only checks for `<@BOT_ID>` or `<@!BOT_ID>` in the message content text. Discord's Reply feature does not inject @mention text into the content — it sets `message_reference` and `referenced_message` fields instead. These fields are not checked.

## Root Cause

`discord.zig` line ~825:
```zig
if (self.require_mention and guild_id != null) {
    const bot_uid = self.bot_user_id orelse "";
    if (!isMentioned(content, bot_uid)) {
        return;  // Silently drops reply-to-bot messages
    }
}
```

## Proposed Fix

Check `referenced_message.author.id` against the bot's user ID as an alternative trigger:

```zig
if (self.require_mention and guild_id != null) {
    const bot_uid = self.bot_user_id orelse "";
    const is_mentioned = isMentioned(content, bot_uid);

    // Check if this is a reply to the bot's own message
    const is_reply_to_bot: bool = if (d_obj.get("referenced_message")) |ref_val| blk: {
        if (ref_val == .object) {
            if (ref_val.object.get("author")) |ref_author| {
                if (ref_author == .object) {
                    if (ref_author.object.get("id")) |ref_id| {
                        if (ref_id == .string) {
                            break :blk std.mem.eql(u8, ref_id.string, bot_uid);
                        }
                    }
                }
            }
        }
        break :blk false;
    } else false;

    if (!is_mentioned and !is_reply_to_bot) {
        return;
    }
}
```

This preserves the existing mention behavior while adding reply detection. No new config keys needed — it's a natural extension of `require_mention`.

## Additional Issues Found

While investigating, also discovered:

1. **`guild_id` must be a JSON string, not integer** — Discord snowflake IDs like `878250798528745542` exceed JSON integer precision. When provided as a number, the entire Discord config block is silently ignored (shows "not configured"). Should either document this or add type coercion.

2. **xAI provider base_url is wrong** — Built-in URL is `https://api.x.ai` but the correct endpoint is `https://api.x.ai/v1`. Results in 404 errors. Workaround: set `"base_url": "https://api.x.ai/v1"` in provider config.

## Environment

- nullclaw v2026.3.4 (Windows 11, Zig 0.15.2)
- Discord channel with `require_mention: true`
- Tested with custom build — fix confirmed working

manelsen:
## Summary
This fixes #337.

When `require_mention: true` is enabled for Discord guild messages, nullclaw previously only treated explicit text mentions like `<@BOT_ID>` as directed input.

That missed a common Discord interaction pattern: replying directly to the bot's message. Discord replies do not inject a text mention into `content`; instead, they carry the target message in `referenced_message`.

## What changed
- keep the existing explicit mention check unchanged
- add a second directed-message condition: the incoming message is a reply to a message authored by the bot
- restrict that behavior to the existing `require_mention` gate for guild messages
- add focused tests for:
  - unmentioned guild messages still being blocked
  - replies to bot messages being accepted
  - replies to non-bot messages still being blocked

## Why this is correct
The `require_mention` invariant is not "the raw text must contain `<@bot>`".
It is "the message must be clearly directed at the bot".

A Discord reply to the bot satisfies that invariant just as strongly as an explicit mention does, so the new behavior broadens the detection signal without weakening the intent check.

## Validation
- `zig build test`

## Notes
Two unrelated local untracked files were intentionally excluded from this PR:
- `docs/hexagonal-functional-core-gap-map.md`
- `resolver.sh`

### DIFF
--- a/src/channels/discord.zig
+++ b/src/channels/discord.zig
@@ -213,6 +213,11 @@ pub const DiscordChannel = struct {
 
     fn isReplyToBot(d_obj: std.json.ObjectMap, bot_user_id: []const u8) bool {
         if (bot_user_id.len == 0) return false;
+        const message_type = d_obj.get("type") orelse return false;
+        switch (message_type) {
+            .integer => |value| if (value != 19) return false,
+            else => return false,
+        }
         const referenced_message = d_obj.get("referenced_message") orelse return false;
         const referenced_obj = switch (referenced_message) {
             .object => |o| o,
@@ -1271,7 +1276,7 @@ test "discord handleMessageCreate require_mention accepts reply to bot message" 
     defer alloc.free(ch.bot_user_id.?);
 
     const msg_json =
-        \\{"d":{"channel_id":"c-2","guild_id":"g-2","content":"reply text","author":{"id":"u-2","bot":false},"referenced_message":{"author":{"id":"bot-1","bot":true}}}}
+        \\{"d":{"channel_id":"c-2","guild_id":"g-2","type":19,"content":"reply text","author":{"id":"u-2","bot":false},"referenced_message":{"author":{"id":"bot-1","bot":true}}}}
     ;
     const parsed = try std.json.parseFromSlice(std.json.Value, alloc, msg_json, .{});
     defer parsed.deinit();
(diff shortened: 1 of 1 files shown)


DonPrus:
@manelsen Tomorrow I’ll try to review all the PRs and get through even the larger ones :)
The bigger PRs are harder to work through, especially when several of them are fixing the same problems.

manelsen:
Thank you, @DonPrus .

I'm working on some "quick wins" today, to rest a little from that big bad refactor. No hurry. There's a lot to do.

PR #384, opened 2026-03-09
fix(streaming): hide pipe-delimited tool control tokens

### LINKED ISSUE #349: tool_calls visible in replys to user
### Description

Image

Image

### Expected behavior

I should not see tool_call_begin when chatting with the bot

### Steps to reproduce

see picture

### Version

2026.3.7

### OS

Linux

manelsen:
## Summary
- extend the streaming tag filter to suppress pipe-delimited tool control blocks
- keep Telegram/Web streaming drafts from showing raw `<|tool_call_begin|>` / `<|tool_result_begin|>` tokens
- add regression coverage for split-chunk control markers

## Problem
Some providers stream tool-control sentinels such as `<|tool_call_begin|>` instead of the XML-style `<tool_call>` tags that nullclaw already strips.

In Telegram streaming mode, those control tokens could leak into the draft message that users see while the response is still in flight.

## Root Cause
`streaming.TagFilter` only recognized two hidden block formats:
- `<tool_call> ... </tool_call>`
- `<tool_result ...> ... </tool_result>`

Provider-specific pipe-delimited control markers were treated as normal text, so they passed through the outbound streaming sink.

## Fix
Teach `TagFilter` to treat these markers as hidden control blocks too:
- `<|tool_call_begin|> ... <|tool_call_end|>`
- `<|tool_result_begin|> ... <|tool_result_end|>`

The filter still works chunk-by-chunk, so partially received markers are suppressed correctly even when the provider splits them across streaming events.

## Why this approach
- fixes the leak in the shared streaming layer instead of patching Telegram-specific presentation code
- covers both Telegram and any other outbound path that uses the same filter
- keeps the invariant simple: tool-control transport tokens must never be forwarded to the user-visible stream

## Testing
- `zig build test`
- added regression tests for:
  - a full `<|tool_call_begin|> ... <|tool_call_end|>` block in one chunk
  - a `<|tool_result_begin|> ... <|tool_result_end|>` block split across chunks

Closes #349.

### DIFF
--- a/src/daemon.zig
+++ b/src/daemon.zig
@@ -669,6 +669,20 @@ fn publishStreamingChunk(ctx_ptr: *anyopaque, event: streaming.Event) void {
     };
 }
 
+fn supportsStreamingOutbound(channel: []const u8) bool {
+    return std.mem.eql(u8, channel, "web") or std.mem.eql(u8, channel, "telegram");
+}
+
+fn makeStreamingSinkForChannel(
+    channel: []const u8,
+    raw_sink: streaming.Sink,
+    filter: *streaming.TagFilter,
+) ?streaming.Sink {
+    if (!supportsStreamingOutbound(channel)) return null;
+    filter.* = streaming.TagFilter.init(raw_sink);
+    return filter.sink();
+}
+
 fn inboundDispatcherThread(
     allocator: std.mem.Allocator,
     event_bus: *bus_mod.Bus,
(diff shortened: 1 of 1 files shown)


juvenn:
Taking a close look at the screenshot, the actual tool call is in the following weird format. The close tag is not xml at all. 🤔 

```xml
<|tool_calls_section_begin|>
    <|tool_call_begin|>
        functions.file_read:0
        <|tool_call_argument_begin|>
            {
                "path": "skills/rmv-transit/SKILL.md"
            }
        <|tool_call_end|>
    <|tool_call_begin|>
        functions.shell:1
        <|tool_call_argument_begin|>
            {
                "command": "ls -la skills/rmv-transit/"
            }
        <|tool_call_end|>
<|tool_calls_section_end|>
```

kimi2

And here's another one, they're not consistent at all.

```xml
<tool_calls_section_begin>
  <tool_call_begin> functions.shell:5<{"command": "ls -la scripts/nullclaw-release-checker/"}</tool_call>
<tool_calls_section_end|>
```

kimi2

Issue #354, opened 2026-03-07
Service stops working after Homebrew upgrade


kronk307:
## Summary

After upgrading nullclaw via Homebrew (e.g. `brew upgrade nullclaw`), the daemon silently stops working. The root cause is that `nullclaw service install` writes the LaunchAgent plist with a hardcoded path to the versioned Cellar binary:

```
/opt/homebrew/Cellar/nullclaw/2026.3.5/bin/nullclaw
```

When Homebrew upgrades to a new version, that path is removed. `launchctl` then fails to launch the process, but reports the service as `running/loaded` — so there's no obvious error message.

## Steps to reproduce

1. Install nullclaw via Homebrew and run `nullclaw service install`
2. Run `brew upgrade nullclaw`
3. Send a message — no response
4. Run `nullclaw service status` — reports `running/loaded` (misleading)
5. Run `ps aux | grep nullclaw` — no process found

## Expected behavior

The plist should reference the Homebrew shim/symlink (`/opt/homebrew/bin/nullclaw`) instead of the versioned Cellar path, so it survives upgrades automatically. Alternatively, `brew upgrade` should trigger a plist reinstall (e.g. via a Homebrew post-install hook).

## Workaround

```sh
nullclaw service uninstall
nullclaw service install
nullclaw service start
```

## Environment

- macOS (darwin)
- Install method: Homebrew
- Old version: 2026.3.5 → New version: 2026.3.7

## Notes
It also worth noting that after upgrading nullclaw, the Screen Recording permissions on Mac reset, probably due a new binary path.

datvance:
I ran into this same problem on linux using homebrew. brew installs to version specific bin: `/home/linuxbrew/.linuxbrew/Cellar/nullclaw/2026.3.8/bin`

manelsen:
Hi @DonPrus — the core issue (versioned Cellar path in the launchd plist breaking after `brew upgrade`) was fixed by **PR #398** (merged 2026-03-10), which introduced `preferredHomebrewShimPath()` in `src/service.zig` to rewrite the path to the stable shim (`/opt/homebrew/bin/nullclaw`). Tests cover Apple Silicon, Intel Mac, and Linux Homebrew.

The secondary comment about macOS permissions resetting after an upgrade is a macOS entitlements/code-signing constraint outside the project's direct control, not a bug in nullclaw.

Would you mind closing this one with a note about the permissions limitation?

manelsen:
[...]

PR #398, opened 2026-03-09
fix(service): use stable Homebrew path for macOS launchd

manelsen:
## Summary
- use the stable Homebrew shim path for launchd plists on macOS
- avoid writing versioned Cellar paths into the LaunchAgent
- add tests for Apple Silicon and Intel Homebrew installs

## Why
`nullclaw service install` could write a versioned Cellar binary path into the plist. After `brew upgrade`, that path disappears and the service silently stops working.

## Validation
- `zig test src/service.zig`

### DIFF
--- a/src/service.zig
+++ b/src/service.zig
@@ -285,7 +285,8 @@ fn preferredHomebrewShimPath(allocator: std.mem.Allocator, exe_path: []const u8)
         return null;
     }
 
-    const candidate = try std.fs.path.join(allocator, &.{ exe_path[0..cellar_index], "bin", "nullclaw" });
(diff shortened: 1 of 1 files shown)


datvance:
This PR is macOS specific but I ran into the same problem on linux today. Should it be a separate PR?
https://github.com/nullclaw/nullclaw/issues/354#issuecomment-4027387854

manelsen:
> This PR is macOS specific but I ran into the same problem on linux today. Should it be a separate PR? [#354 (comment)](https://github.com/nullclaw/nullclaw/issues/354#issuecomment-4027387854)

If the PR doesn't work for you on Linux, open a new issue.

Issue #434, opened 2026-03-11
Multi-Agent setup should have topic isolation (message_thread_id) for Telegram


ats-bcon:
### Description

Dont know if its a bug or a feature, but **in a multi-agent setup: all messages from the same Telegram group, regardless of which topic they're in, get the same session + same agent. There is zero topic isolation today.**

Suggestion / Expectation: In a multi agent setup it should be possible to configure the topicId ( message_thread_id ) so that the dedicated agent adheres to that specific topic. 

Following is an (example) of a multi-agent setup that I refer to:

_"agents": {
		"defaults": {
			"model": {
				"primary": "openrouter/moonshotai/kimi-k2.5"
			}
		},
		"list": [
			{
				"id": "orchestrator",
				"provider": "openrouter",
				"model": "openrouter/moonshotai/kimi-k2.5",
				"temperature": 0.4,
				"system_prompt": "You are the Orchestrator. Your job is to manage tasks and delegate them."
			},
			{
				"id": "coder",
				"provider": "openrouter",
				"model": "openrouter/moonshotai/kimi-k2.5",
				"temperature": 0.2,
				"system_prompt": "You are an expert software engineer. Focus on clean, efficient code."
			},
			{
				"id": "translator",
				"provider": "openrouter",
				"model": "openrouter/moonshotai/kimi-k2.5",
				"temperature": 0.6,
				"system_prompt": "You are a professional translator. Provide accurate and natural translations."
			}
		]
	},
	"bindings": [
		{
			"agent_id": "orchestrator",
			"match": {
				"channel": "telegram",
				"peer": {
					"kind": "direct",
					"id": "XXXXXXXX"
				}
			}
		},
		{
			"agent_id": "coder",
			"match": {
				"channel": "telegram",
				"peer": {
					"kind": "group",
					"id": "-100YYYYYYYYY"
				}
			}
		},
		{
			"agent_id": "translator",
			"match": {
				"channel": "telegram",
				"peer": {
					"kind": "group",
					"id": "-100YYYYYYYYY"
				}
			}
		}
	],_

vedmalex:
Implemented in PR #450: https://github.com/nullclaw/nullclaw/pull/450

This adds a documented config pattern for Telegram forum-topic agent routing:

- use `bindings[].match.peer.id = "<chat_id>:thread:<topic_id>"` for a topic-specific agent
- keep `bindings[].match.peer.id = "<chat_id>"` as the fallback agent for the rest of the group/forum

[...]

PR #436, opened 2026-03-11
fix: harden Telegram forum topics and configurable command menus

### LINKED ISSUE #435: Fix Telegram forum topics, task reactions, and configurable command menus
## Summary

The Telegram channel still had several gaps around forum topics, task-state visibility, and operator ergonomics.

In practice this showed up as a mix of routing bugs and missing UX:

- forum topic messages could lose topic isolation and fall back to the base group session
- legacy topic sessions were stored under a non-canonical key format and could split history after runtime changes
- the bot could not expose topic/session state clearly inside Telegram
- task lifecycle visibility on the user's triggering message was missing or incomplete
- the Telegram slash-command menu was static and not configurable per deployment or chat type

These issues make Telegram forum workflows harder to operate and increase the risk of cross-topic context bleed.

### 1. Forum-topic session routing was not robust enough

Telegram forum traffic can arrive with `message_thread_id`, but some topic updates may only expose the thread indirectly via a topic reply root.

Before this fix, the runtime relied too narrowly on the top-level thread field, which meant topic traffic could be treated as plain group traffic and land in the base chat session instead of an isolated per-topic session.

### 2. Topic session keys were inconsistent

Legacy topic sessions used the older `#topic:<id>` form, while the newer runtime direction is the canonical `:thread:<id>` form.

Without migration, the same logical topic could end up split across:

- an older legacy session key
- a newer canonical session key

That risks fragmented transcript history, usage accounting, and session-scoped memory.

### 3. Telegram topic operations were incomplete

The channel did not have a full built-in workflow for:

- creating a new forum topic from Telegram itself
- seeing which in-memory sessions currently exist for the active forum chat
- consistently returning replies, typing indicators, and other Telegram actions back into the same topic thread

### 4. Task-state visibility inside Telegram was weak

There was no configurable lifecycle feedback on the triggering user message for:

- accepted
- running
- done
- failed

That made it hard to see whether the bot had accepted the task, was still working, or had already finished.

### 5. The command menu was static and not deployment-friendly

The Telegram slash-command menu was a single static payload.

This caused several problems:

- no way to disable menu publication entirely
- no way to publish a different menu for private chats vs group chats
- no way to hide topic-only commands from private chats
- no grouped operator-facing entry point such as `/menu`

## Root Cause

This was not one isolated bug. The problem was spread across several layers:

- Telegram update ingress did not infer thread identity from topic reply roots when the top-level field was absent
- Telegram API helpers did not consistently support topic-aware actions and forum methods
- session management did not provide a safe migration path from legacy topic keys to canonical thread keys
- Telegram runtime surfaces did not provide built-in topic map or lifecycle reaction workflows
- command publication logic was hard-coded instead of config-driven

## Expected Behavior

After the fix, Telegram should behave as follows:

1. Messages in a forum topic must resolve to an isolated per-topic session.
2. Existing legacy `#topic:<id>` sessions must be migrated forward to canonical `:thread:<id>` keys without dropping transcript history, usage, or session-scoped memory.
3. Telegram-specific built-ins should support:
   - `/topic <name>` to create a new forum topic
   - `/topics` and `/topic-map` to inspect active topic/session state for the current chat
4. Task lifecycle reactions should be configurable and optional.
5. Telegram command menu publication should be configurable with modes:
   - `off`
   - `flat`
   - `scoped`
6. In `scoped` mode, private chats and group chats should receive different command menus.
7. `/menu` should provide a grouped operator-facing help surface.

## Acceptance Criteria

- Topic updates are routed to topic-specific sessions instead of the base group session.
- The canonical session key format is `:thread:<id>`.
- Legacy `#topic:<id>` sessions are migrated best-effort into the canonical key when safe.
- Transcript history and session-scoped memory survive the migration path.
- Topic-aware Telegram sends, replies, typing indicators, and file sends stay in the correct thread.
- `/topic`, `/topics`, and `/topic-map` are available behind Telegram config flags.
- Task lifecycle reactions can be enabled or disabled in config and use configurable emoji per state.
- Telegram command menu publication is controlled from config and supports scoped publication for private vs group chats.
- Tests cover the new routing, migration, parsing, and command-menu behavior.

vedmalex:
## Summary

Fixes #435.

This PR closes a Telegram-specific gap that spanned three related areas:

- forum topic routing and session isolation
- task lifecycle visibility inside Telegram
- command menu publication and operator ergonomics

Before this change, Telegram forum workflows were functional only partially. Topic traffic could lose its thread identity in some cases, older topic session keys could fragment history, task-state visibility was weak, and the slash-command menu was static instead of deployment-configurable.

### 1. Hardened forum-topic routing and canonical session keys

- resolve Telegram group/forum sessions through a canonical per-thread key using `:thread:<id>`
- add best-effort migration from legacy `#topic:<id>` session keys into canonical thread keys
- migrate live in-memory sessions, persisted transcript history, usage counters, and session-scoped memory when safe
- add session snapshot support so Telegram-facing status surfaces can inspect active sessions

### 2. Improved Telegram update ingress and topic-aware runtime behavior

- parse and propagate `message_thread_id` from Telegram updates
- add fallback inference for topic replies when Telegram only exposes the topic via the reply root
- ensure topic-aware sends, typing indicators, and media/file paths stay inside the correct thread
- add Telegram built-ins for topic workflows:
  - `/topic <name>` to create a forum topic
  - `/topics` and `/topic-map` to inspect active topic/session state for the current chat

### 3. Added configurable task lifecycle reactions

- add Telegram lifecycle reactions for:
  - `accepted`
  - `running`
  - `done`
  - `failed`
- make reactions configurable per Telegram account
- allow emoji override per lifecycle state
- keep the feature best-effort so Telegram reaction failures do not break the bot's normal reply flow

### 4. Added configurable Telegram command-menu publication

- replace the previous static Telegram command payload with config-driven publication
- add `commands_menu_mode` to Telegram config with modes:
  - `off`
  - `flat`
  - `scoped`
- in `scoped` mode, publish different command menus for:
  - `all_private_chats`
  - `all_group_chats`
- keep topic-only commands out of private chat menus
- add `/menu` as a grouped help alias for operator-facing command discovery
- update help output so commands are grouped by workflow instead of presented as a single flat block

### 5. Added test coverage

This PR adds or updates coverage for:

- Telegram update ingress thread parsing and fallback behavior
- canonical/legacy session migration behavior
- topic/session map reporting surfaces
- grouped help and `/menu` handling
- Telegram command menu JSON generation and scoped delete payloads
- Telegram config parsing for reactions, topic commands, and menu modes

## Why This Fix Matters

Telegram forum chats are one of the most important multi-session surfaces for operating the bot in practice.

Without robust topic routing, users can accidentally mix context between different discussion threads. Without lifecycle reactions and topic/session visibility, it is harder to see what the bot is doing. Without configurable menus, the Telegram UX becomes noisy in some contexts and incomplete in others.

This change makes the Telegram runtime safer, easier to operate, and closer to how forum-based workflows behave in real use.

### Automated

  - passed: `5244/5255`
  - skipped: `11`
  - success

### Live verification

Validated against a real forum-enabled Telegram supergroup with the current bot deployment:

- forum topic creation works through `/topic`
- topic replies stay in the correct topic thread
- canonical `:thread:<id>` session routing is used end-to-end
- lifecycle reactions are applied on the triggering user messages
- scoped command publication is live via Telegram Bot API:
  - default scope cleared
  - private-chat scope published without topic-only commands
  - group-chat scope published with topic-only commands

## Notes

- Local runtime config was updated only on the deployment machine; no repo-local secret config is included in this PR.
- Reaction and menu publication remain configurable per Telegram account so deployments can opt into stricter or quieter behavior as needed.

### DIFF
--- a/src/agent/dispatcher.zig
+++ b/src/agent/dispatcher.zig
@@ -634,9 +634,137 @@ fn stripTrailingXml(input: []const u8) []const u8 {
     return std.mem.trimRight(u8, trimmed_right[0..start_tag], " \t\r\n");
 }
 
+fn isToolNameChar(c: u8) bool {
+    return switch (c) {
+        'a'...'z', 'A'...'Z', '0'...'9', '_', '-', '.' => true,
+        else => false,
+    };
+}
+
+fn isPlausibleToolName(name: []const u8) bool {
+    if (name.len == 0) return false;
+    for (name) |c| {
+        if (!isToolNameChar(c)) return false;
+    }
+    return true;
+}
+
+fn hasMalformedQuotedColonToolName(raw_json: []const u8) bool {
+    const name_key = "\"name\"";
+    const name_idx = std.mem.indexOf(u8, raw_json, name_key) orelse return false;
+
+    var remaining = std.mem.trimLeft(u8, raw_json[name_idx + name_key.len ..], " \t\r\n");
+    if (remaining.len == 0 or remaining[0] != ':') return false;
+
+    remaining = std.mem.trimLeft(u8, remaining[1..], " \t\r\n");
+    if (remaining.len < 2 or remaining[0] != '"') return false;
+
+    remaining = remaining[1..];
+    if (remaining.len == 0 or remaining[0] != ':') return false;
+
+    remaining = std.mem.trimLeft(u8, remaining[1..], " \t\r\n");
+    return remaining.len > 0 and remaining[0] == '"';
+}
+
+fn recoverToolNameFromRawJson(raw_json: []const u8) ?[]const u8 {
+    const name_key = "\"name\"";
+    const arguments_key = "\"arguments\"";
+
+    const name_idx = std.mem.indexOf(u8, raw_json, name_key) orelse return null;
+    const after_name = raw_json[name_idx + name_key.len ..];
+    const colon_idx = std.mem.indexOfScalar(u8, after_name, ':') orelse return null;
+    const after_colon = after_name[colon_idx + 1 ..];
+
+    const window = if (std.mem.indexOf(u8, after_colon, arguments_key)) |args_idx|
+        after_colon[0..args_idx]
+    else
+        after_colon;
+
+    var i: usize = 0;
+    while (i < window.len) : (i += 1) {
+        if (!isToolNameChar(window[i])) continue;
+
+        const start = i;
+        var end = i + 1;
+        while (end < window.len and isToolNameChar(window[end])) : (end += 1) {}
+
+        const candidate = window[start..end];
+        if (!std.mem.eql(u8, candidate, "name") and !std.mem.eql(u8, candidate, "arguments")) {
+            return candidate;
+        }
+
+        i = end;
+    }
+
+    return null;
+}
+
+fn extractRawArgumentsJson(raw_json: []const u8) ?[]const u8 {
+    const arguments_key = "\"arguments\"";
+    const args_idx = std.mem.indexOf(u8, raw_json, arguments_key) orelse return null;
+    const after_args = raw_json[args_idx + arguments_key.len ..];
+    const colon_idx = std.mem.indexOfScalar(u8, after_args, ':') orelse return null;
+    return extractJsonObject(after_args[colon_idx + 1 ..]);
+}
+
+fn normalizeSalvagedArgumentsJson(allocator: std.mem.Allocator, raw_json: []const u8) ![]u8 {
+    const args_src = extractRawArgumentsJson(raw_json) orelse return allocator.dupe(u8, "{}");
+    var parsed_args = std.json.parseFromSlice(std.json.Value, allocator, args_src, .{}) catch blk: {
+        const repaired = repairJson(allocator, args_src) catch return allocator.dupe(u8, "{}");
+        defer allocator.free(repaired);
+        break :blk std.json.parseFromSlice(std.json.Value, allocator, repaired, .{}) catch return allocator.dupe(u8, "{}");
+    };
+    defer parsed_args.deinit();
+
+    return try std.json.Stringify.valueAlloc(allocator, parsed_args.value, .{});
+}
+
+fn salvageMalformedToolCallJson(allocator: std.mem.Allocator, raw_json: []const u8) !?ParsedToolCall {
+    if (!hasMalformedQuotedColonToolName(raw_json)) return null;
+
+    const recovered_name = recoverToolNameFromRawJson(raw_json) orelse return null;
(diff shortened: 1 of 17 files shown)


ats-bcon:
Thanks, cant wait for the merge!

ats-bcon:
hi @vedmalex great PR. I am a bit confused how to setup agents and topics in the config.json. I had opened this issue, it seems very related to your PR https://github.com/nullclaw/nullclaw/issues/434. How can the "topic_id"s be added in the config to the agents? (I think nullclaw needs a better documentation on this issue)

vedmalex:
Follow-up is now in PR #450: https://github.com/nullclaw/nullclaw/pull/450

Short answer: you do **not** add a separate `topic_id` field under `channels.telegram`.

For Telegram forum topics, the topic-specific agent is configured in `bindings` using the peer id form:

```json
{
  "agent_id": "coder",
  "match": {
    "channel": "telegram",
    "account_id": "main",
    "peer": { "kind": "group", "id": "-1001234567890:thread:42" }
  }
}
```

Then keep a plain group binding as fallback for the rest of the forum:

```json
{
  "agent_id": "orchestrator",
  "match": {
    "channel": "telegram",
    "account_id": "main",
    "peer": { "kind": "group", "id": "-1001234567890" }
  }
}
```

So:
- topic `42` routes to `coder`
- the rest of the same Telegram forum falls back to `orchestrator`

PR #450 adds both the routing support and explicit documentation for this config pattern.

vedmalex:
`account_id` is the id of the configured **Telegram account entry**. It is not a topic id and not an agent id.

How it works:

- In the normal config shape, the key under `channels.telegram.accounts` becomes the account id.
- So `accounts.main` means `account_id = "main"`.
- If you had another Telegram bot account like `accounts.sub`, that would mean `account_id = "sub"`.

What it is used for:

- In `bindings`, `match.account_id` scopes the binding to one specific Telegram account.
- If `match.account_id` is omitted, the binding can match any Telegram account for that channel.

Effect on Telegram delivery:

- Incoming updates are handled by the Telegram account that received them.
- Routing uses that same `account_id`.
- Replies go back out through the same Telegram account/runtime that handled the message.

So if one binding uses `account_id = "main"` and another uses `account_id = "sub"`, they apply to different configured Telegram bot accounts. This does **not** split one single Telegram account's traffic by itself.

You usually only care about `account_id` when the same nullclaw instance runs multiple Telegram bot tokens/accounts. In a normal single-bot setup, `main` is enough.

I also added explicit docs for this in PR #450, together with the topic-binding docs.

PR #442, opened 2026-03-12
feat(tools): implement Hashline for precise and verifiable code editing

manelsen:
### feat(tools): implement Hashline for precise and verifiable code editing

This PR introduces **Hashline**, a new pair of tools (`file_read_hashed` and `file_edit_hashed`) designed to solve the **Harness Problem** in AI-assisted coding. 

Traditional string-replacement tools are fragile. They often fail due to minor whitespace discrepancies, indentation mismatches, or stale file states (when the file changes on disk while the LLM is "thinking"). Hashline addresses these issues by using **Line Tagging with Content Hashing**.

#### How it works:

1.  **`file_read_hashed`**: Instead of raw text, it returns lines prefixed with a stable identifier: `L<line_number>:<hash>| <content>`.
    *   The hash is generated based on the **normalized content** (ignoring leading/trailing whitespace), making it resilient to formatting variations.
    *   Example: `L10:a3f| const std = @import("std");`

2.  **`file_edit_hashed`**: When the LLM wants to edit a file, it references these tags.
    *   **Deterministic Validation**: The tool verifies if the current line in the file matches the hash provided by the LLM. If the hashes don't match (indicating a dirty state or a wrong line), the edit is rejected.
    *   **Atomic Range Replacement**: Supports replacing blocks of code by specifying a `start_target` and an optional `end_target`.

#### Why this improves NullClaw:

*   **Zero Indentation Errors**: Since the anchor is a hash of the content, the LLM no longer needs to perfectly replicate the indentation of the "old text" to find the match.
*   **Reduced Token Usage**: The LLM doesn't need to provide large chunks of "context" lines to disambiguate where to apply the change. It just points to the ID.
*   **Safety & Integrity**: Prevents "hallucinated" edits or accidental corruption of files when the workspace state changes mid-session.
*   **Subagent Robustness**: Crucial for autonomous subagents that perform complex, multi-step refactoring tasks.

#### Changes:
- Added `src/tools/file_read_hashed.zig`
- Added `src/tools/file_edit_hashed.zig`
- Registered tools in `src/tools/root.zig`
- Updated test suites and verified zero memory leaks.

### DIFF
--- a/src/agent/root.zig
+++ b/src/agent/root.zig
@@ -2269,7 +2269,9 @@ pub const Agent = struct {
     }
 
     fn tool_call_updates_tools_md(allocator: std.mem.Allocator, call: ParsedToolCall) bool {
-        if (!std.mem.eql(u8, call.name, "file_write") and !std.mem.eql(u8, call.name, "file_edit")) return false;
+        if (!std.mem.eql(u8, call.name, "file_write") and
+            !std.mem.eql(u8, call.name, "file_edit") and
+            !std.mem.eql(u8, call.name, "file_edit_hashed")) return false;
 
         const parsed = std.json.parseFromSlice(std.json.Value, allocator, call.arguments_json, .{}) catch return false;
         defer parsed.deinit();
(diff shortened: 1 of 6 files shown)


manelsen:
Thank you so much @can1357 .

DonPrus:
Great job, @manelsen ! But try not to add commits from other branches :)

manelsen:
> Great job, @manelsen ! But try not to add commits from other branches :)

Thank you, @DonPrus . I'll be mindful of that next time.

PR #450, opened 2026-03-12
fix: support Telegram topic-specific agent bindings

### LINKED ISSUE #434: Multi-Agent setup should have topic isolation (message_thread_id) for Telegram

vedmalex:
## Summary

Fixes #434.

This follow-up adds the missing piece that came up during review of #436:

- topic isolation already existed at the session level
- but config-level agent routing for specific Telegram forum topics was still unclear and not explicitly supported as a first-class binding pattern

This PR makes Telegram forum topic routing usable in multi-agent setups and documents the exact config format.

## What Changed

- extend Telegram route resolution so forum topics can match a more specific peer id in the form `"<chat_id>:thread:<topic_id>"`
- pass the base Telegram group as `parent_peer` during topic routing so plain group bindings still act as fallback
- keep canonical session storage unchanged: topic conversations still persist under `:thread:<id>` session keys
- add documentation to `README.md` and `docs/en/configuration.md` clarifying that:
  - there is no separate `topic_id` field under `channels.telegram`
  - topic-specific routing is configured in `bindings`
  - the correct peer id format is `"<chat_id>:thread:<topic_id>"`
  - plain group bindings continue to serve as fallback for all other topics

## Example

```json
{
  "bindings": [
    {
      "agent_id": "coder",
      "match": {
        "channel": "telegram",
        "account_id": "main",
        "peer": { "kind": "group", "id": "-1001234567890:thread:42" }
      }
    },
    {
      "agent_id": "orchestrator",
      "match": {
        "channel": "telegram",
        "account_id": "main",
        "peer": { "kind": "group", "id": "-1001234567890" }
      }
    }
  ]
}
```

With that config:

- topic `42` routes to `coder`
- the rest of the forum falls back to `orchestrator`

## Validation

  - passed: `5301/5312`
  - skipped: `11`
  - success

### DIFF
--- a/src/agent_bindings_config.zig
+++ b/src/agent_bindings_config.zig
@@ -279,8 +279,9 @@ pub fn applyBindingUpdate(
             }
         }
 
-        const desired = try buildManagedBinding(allocator, target, normalized_agent_id);
-        errdefer freeBinding(allocator, &desired);
+        var desired = try buildManagedBinding(allocator, target, normalized_agent_id);
+        var desired_owned = true;
+        errdefer if (desired_owned) freeBinding(allocator, &desired);
 
         var next: std.ArrayListUnmanaged(agent_routing.AgentBinding) = .empty;
         errdefer freeBindingList(allocator, &next);
@@ -289,12 +290,14 @@ pub fn applyBindingUpdate(
         for (cfg.agent_bindings, 0..) |binding, idx| {
             if (existing != null and idx == existing.?.index) {
                 try next.append(allocator, desired);
+                desired_owned = false;
                 continue;
             }
             try next.append(allocator, try dupBinding(allocator, binding));
         }
         if (existing == null) {
             try next.append(allocator, desired);
+            desired_owned = false;
         }
 
         replaceOwnedBindings(allocator, cfg, try next.toOwnedSlice(allocator));
@@ -347,7 +350,9 @@ pub fn persistBindingUpdate(
 
     var cfg = try loadConfigFromPath(arena, config_path);
     const result = try applyBindingUpdate(arena, &cfg, target, requested_agent);
-    try cfg.save();
+    if (result.status != .unchanged) {
+        try cfg.save();
+    }
     return result;
 }
 
@@ -517,4 +522,70 @@ test "applyBindingUpdate clear keeps inherited unscoped peer binding" {
     try std.testing.expectEqual(@as(usize, 1), cfg.agent_bindings.len);
     try std.testing.expectEqualStrings("coder-agent", cfg.agent_bindings[0].agent_id);
     try std.testing.expect(!cfg.agent_bindings_runtime_owned);
+}
+
+test "persistBindingUpdate skips rewriting config when binding is unchanged" {
+    const allocator = std.testing.allocator;
+    var tmp = std.testing.tmpDir(.{});
+    defer tmp.cleanup();
+
+    const base = try tmp.dir.realpathAlloc(allocator, ".");
+    defer allocator.free(base);
+    const config_path = try std.fmt.allocPrint(allocator, "{s}/config.json", .{base});
+    defer allocator.free(config_path);
+
+    const initial_content =
+        \\{
+        \\  "agents": {
+        \\    "defaults": {
+        \\      "model": {
+        \\        "primary": "ollama/qwen2.5-coder:14b"
+        \\      }
+        \\    },
+        \\    "list": [
+        \\      {
+        \\        "id": "Coder Agent",
+        \\        "provider": "ollama",
+        \\        "model": "qwen2.5-coder:14b"
+        \\      }
+        \\    ]
+        \\  },
+        \\  "bindings": [
+        \\    {
+        \\      "agent_id": "coder-agent",
+        \\      "comment": "Managed by Telegram /bind",
+        \\      "match": {
+        \\        "channel": "telegram",
+        \\        "account_id": "main",
+        \\        "peer": {
+        \\          "kind": "group",
+        \\          "id": "-100123:thread:42"
+        \\        }
+        \\      }
+        \\    }
+        \\  ]
+        \\}
+    ;
+
+    {
+        const file = try std.fs.createFileAbsolute(config_path, .{});
+        defer file.close();
+        try file.writeAll(initial_content);
+    }
+
+    const result = try persistBindingUpdate(allocator, config_path, .{
+        .channel = "telegram",
+        .account_id = "main",
+        .peer = .{ .kind = .group, .id = "-100123:thread:42" },
+        .comment = "Managed by Telegram /bind",
+    }, "Coder Agent");
+
+    try std.testing.expect(result.status == .unchanged);
+
+    const file = try std.fs.openFileAbsolute(config_path, .{});
+    defer file.close();
+    const persisted = try file.readToEndAlloc(allocator, 128 * 1024);
+    defer allocator.free(persisted);
+
+    try std.testing.expectEqualStrings(initial_content, persisted);
 }
--- a/src/agent_routing.zig
+++ b/src/agent_routing.zig
@@ -282,6 +282,39 @@ fn hasMatchingRole(binding_roles: []const []const u8, member_roles: []const []co
     return false;
 }
 
+fn firstMatchingBinding(
+    candidates: []const AgentBinding,
+    input: RouteInput,
+    check_peer: ?PeerRef,
+    matched_by: MatchedBy,
+) ?AgentBinding {
+    // Within the same routing tier, prefer bindings scoped to the current
+    // account before broader unscoped fallbacks.
+    for ([_]bool{ true, false }) |prefer_account_scoped| {
+        for (candidates) |binding| {
+            if (prefer_account_scoped != (binding.match.account_id != null)) continue;
+
+            const matches = switch (matched_by) {
+                .peer, .parent_peer => binding.match.peer != null and allConstraintsMatch(binding, input, check_peer),
+                .guild_roles => binding.match.guild_id != null and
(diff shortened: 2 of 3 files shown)


ats-bcon:
@vedmalex , thanks for clarification. I understand now. Very very good that you added documentation, I think other people will ask themselves the same questions.

vedmalex:
Follow-up changes are now pushed in `774e396`.

What changed:
- `/bind` and `/bind clear` are now exact-only for the current Telegram account/chat/topic. They no longer rewrite or remove inherited broader fallback bindings.
- `/bind status` now distinguishes `Exact binding` from `Inherited peer binding`, so the effective route is easier to understand.
- Live hot-apply for `/bind` now cleans up replaced runtime-owned binding slices instead of accumulating old `agent_bindings` in memory.
- Docs were expanded with a minimal end-to-end Telegram example (`agents.list` + `channels.telegram.accounts.main` + fallback binding), plus notes about persistence to `~/.nullclaw/config.json`, exact vs inherited bindings, and binding priority.

Validation:
- `zig build test --summary all` -> `5316/5327 passed`, `11 skipped`
- `zig build -Doptimize=ReleaseSmall` -> passed

This should address the usability/documentation concerns raised earlier and make `/bind` behavior more predictable for operators.

ats-bcon:
I love how your documentation is precise ,cear and easy to understand. Thanks!

ats-bcon:
@vedmalex after the update that includes your PR I get the following crash after "/new". I am not sure, but it is possible it is related to your fix. Maybe some error is not handled gracefully?

grafik

grafik

vedmalex:
## Post-merge regression identified and fixed

@ats-bcon the crash/hang after `/new` you reported is caused by a **dangling provider pointer** in `getOrCreate()`.

### Root cause

When a named-agent session is created (via `/bind <agent>`), `getOrCreate()` calls `holder.provider()` which returns a vtable interface with `ptr` pointing into the stack-local `SessionProviderContext.holder`. The holder is later moved to `session.provider_holder` via struct assignment, but the agent's provider `ptr` still points at the old (now-invalid) location.

This means any message sent to the bound agent goes through a dangling pointer — the provider call either silently fails (no response, ⚡ reaction stays) or crashes on `/new` when trying to reset the session.

### Fix

PR #500 reorders the initialization so the `ProviderHolder` is moved to session-owned memory **before** `holder.provider()` is called. The vtable `ptr` is stable from the start.

Live-tested: `/bind agent` → messages get responses, `/new` works correctly.

ats-bcon:
Great.
Thanks for quick the reply @vedmalex.

 There is one more thing you could clarify to me, once I have configured the 3 agents and the bindings in the config. How do I start sub agents? For instance the coder Agent? I want to request that a bash script is generated in the main chat. The orchestrator should start the coder subagnet and all the talk the coder generates should go to the „Coder“ topic. Would that work? Or am I understanding the whole sub agent game wrongly?

vedmalex:
@ats-bcon good question — let me clarify how the multi-agent + topic setup actually works, because it's a bit different from what you're imagining.

### What topic bindings do

Topic bindings are a **static dispatch** mechanism: each Telegram forum topic is routed to a specific agent profile. When you send a message in a topic, the binding determines which agent (model, system prompt, temperature, etc.) handles it. Each topic gets its own **independent session** with its own conversation history.

### How to use it

You talk to each agent **directly in its own topic**:

- Open the **General** topic → talk to the Orchestrator
- Open the **Coder** topic → ask for a bash script → the Coder agent responds right there
- Open the **Reviewer** topic → paste code for review → the Reviewer agent responds

Each agent has its own context and memory, scoped to its topic.

### What about automatic cross-topic delegation?

The scenario you described — "ask the orchestrator in the main chat, it starts the coder subagent, and the coder's output goes to the Coder topic" — is **not how it works currently**.

Here's what's available:

| Mechanism | What it does | Where output goes |
|-----------|-------------|-------------------|
| **Direct topic message** | You write in the Coder topic | Coder agent replies in the same topic |
| **`delegate` tool** | Orchestrator calls another agent's model synchronously | Result appears **inline** in the orchestrator's reply (same topic) |
| **`spawn` tool** | Orchestrator starts a background task | Result comes back to the orchestrator's own session |

So if the orchestrator uses `delegate` to ask the coder model to generate a script, the script appears as part of the orchestrator's response — **in the same topic where you asked**, not in the Coder topic.

### TL;DR

Think of it as **separate desks, not a team hierarchy**. You walk up to whichever desk (topic) you need. The agents don't pass work between topics automatically — you direct the conversation by choosing which topic to write in.

Hope that clarifies things!

vedmalex:
@ats-bcon one more thing — the scenario you're describing (orchestrator delegates to specialized agents, results routed to specific topics) is actually the domain of two companion projects in the nullclaw ecosystem:

### [nullboiler](https://github.com/nullclaw/nullboiler) — DAG Workflow Orchestrator

This is the orchestration layer that sits **above** nullclaw. It provides:

- **14 step types** including `task`, `fan_out`, `map`, `reduce`, `debate` (N agents argue, then a judge synthesizes), `group_chat` (multi-round conversation between agents with different roles), `router`, `approval` (human-in-the-loop gates), and more
- **Tag-based worker routing**: each step specifies `worker_tags` (e.g. `["coder"]`, `["reviewer"]`), and nullboiler dispatches to the right nullclaw worker
- **Worker handoff**: an agent can return `handoff_to` in its output to chain-delegate to another worker
- **Dispatch protocols**: HTTP webhook, MQTT, Redis Streams

### [nulltickets](https://github.com/nullclaw/nulltickets) — Headless Task Tracker

The durable state layer:

- Pipeline FSM with **role-based stage routing** (each stage has an `agent_role` — when a task transitions, the next role's agent claims it)
- Lease-based ownership, task dependencies (DAG), quality gates
- Queue metrics for orchestrators (`/ops/queue`)

### The architecture

```
nulltickets (what needs doing)  →  nullboiler (who does what, when)  →  nullclaw (does the work)
```

Your workflow would be something like:

```json
{
  "steps": [
    {"id": "plan", "type": "task", "worker_tags": ["orchestrator"], "prompt": "..."},
    {"id": "code", "type": "task", "worker_tags": ["coder"], "depends_on": ["plan"], "prompt": "{{steps.plan.output}}"},
    {"id": "review", "type": "task", "worker_tags": ["reviewer"], "depends_on": ["code"]}
  ]
}
```

Each worker is a nullclaw instance (or gateway endpoint) with specific agent config. The piece that's currently missing is **bridging nullboiler step results back to specific Telegram topics** — that integration doesn't exist yet but is a natural next step.

Hope this points you in the right direction!

ats-bcon:
Thanks, thanks, thanks. 

Your ability to present things clearly is astonishing. You should become the head of documentation here. 
I don’t know if the introductory text you wrote is already somewhere in the documents (or I might have missed it) - but that SHOULD be placed in the documentation: the fact that nullclaw is rather separate „desks“ and not a team ( great metaphor) and that those other project parts handle teaming. That’s would help onboard people on the project a lot, imho. The importance of documentation is very underrated in most projects. 

@vedmalex do you mind putting this content into the docs somewhere?

vedmalex:
@ats-bcon thanks for the kind words! 🙏 You're absolutely right about documentation — that "separate desks vs team" framing is missing from the current docs.

I actually just documented the full integration analysis in `docs/integration-analysis.md` and `docs/integration-roadmap.md` in the repo. The key gaps for your use case are:

1. **Worker endpoint** — nullclaw needs `POST /api/worker` to accept tasks from nullboiler
2. **Send API** — `POST /api/send` to push results to specific Telegram topics
3. **Callback receiver** — `POST /api/callback` to receive nullboiler completion events

Once those 3 endpoints are implemented, your workflow (orchestrator → nullboiler → coder → result in Coder topic) will work end-to-end.

Would you be interested in contributing to the documentation? The "separate desks" metaphor + the architecture diagram would be perfect for a "Multi-Agent Workflows" section in the docs. Or I can add it when I get to the documentation PR.

Either way, glad this helped clarify things!

PR #455, opened 2026-03-12
Add openclaw trader daemon 11155219319101910741

danielndhlovu:
New


DonPrus:
Interesting direction. I do not think this belongs in `nullclaw` core, but I do think there is a solid project here and a lot of the work in this PR can be reused almost directly in a better boundary.

What this PR is adding is effectively a separate trading application: Binance integration, a trading daemon, backtesting/hyperopt loops, and an operator dashboard, plus a large vendored websocket dependency. That is a much better fit for an out-of-tree MCP app/server than for the main nullclaw binary.

The main reasons are architectural:

- this is exchange-specific and domain-specific logic, not core assistant/runtime functionality
- it grows the binary and dependency surface a lot
- it increases the blast radius for config, security, and maintenance
- it couples nullclaw releases to a fast-moving trading app

A practical migration path would be:

1. Move the trading domain code into a separate repo/app
   - `exchange.zig`
   - `market.zig`
   - `strategy.zig`
   - `risk.zig`
   - `execution.zig`
   - `backtest.zig`
   - `hyperopt.zig`
   - indicator modules

   Those files are mostly domain logic already, so they do not need to live in `nullclaw` core.

2. Split the app into 2-3 entrypoints/subcommands
   - `trader daemon` -> owns websocket/market polling, positions, PnL, strategy loop
   - `trader web` -> serves the dashboard/UI
   - `trader mcp` -> exposes a thin MCP tool layer for the assistant

   This keeps the long-running exchange/web concerns inside the trader app, not inside the nullclaw binary.

3. Use MCP as the integration boundary

   `nullclaw` already supports both stdio and HTTP MCP servers and automatically registers remote tools as normal tools with names like:

   - `mcp_trader_get_market_snapshot`
   - `mcp_trader_backtest_strategy`
   - `mcp_trader_place_order`

   So the assistant can call them without nullclaw needing any Binance-specific code.

A good first MCP tool surface would be:

- read-only tools first
  - `get_market_snapshot(symbol, interval, limit)`
  - `get_positions()`
  - `get_open_orders()`
  - `get_strategy_status()`
  - `backtest_strategy(symbol, timeframe, params, start, end)`
  - `run_hyperopt(symbol, timeframe, search_budget)`
- write tools after that
  - `paper_place_order(...)`
  - `cancel_order(...)`
  - `set_risk_limits(...)`
- only then consider live trading tools
  - `place_order(...)`

I would strongly recommend making live order placement a separate explicit tool from paper trading, rather than a runtime flag hidden inside one tool.

For local development, the nullclaw side can be as small as:

```json
{
  "mcp_servers": {
    "trader": {
      "transport": "stdio",
      "command": "/opt/openclaw-trader/bin/trader",
      "args": ["mcp", "--config", "/etc/openclaw-trader/config.json"],
      "env": {
        "BINANCE_API_KEY": "${BINANCE_API_KEY}",
        "BINANCE_API_SECRET": "${BINANCE_API_SECRET}"
      }
    }
  }
}
```

If the trader app runs remotely, then expose an HTTPS MCP endpoint and use:

```json
{
  "mcp_servers": {
    "trader": {
      "transport": "http",
      "url": "https://trader.example.com/mcp",
      "timeout_ms": 10000,
      "headers": {
        "Authorization": "Bearer ${TRADER_MCP_TOKEN}"
      }
    }
  }
}
```

Another detail that would make the split cleaner: the web dashboard should talk to the trader app directly, not to nullclaw. The dashboard is really an operator UI for the trading system, not part of the assistant runtime.

So from my side this is not “the idea is wrong”, it is more “the integration boundary is in the wrong place”. I think this would work much better as a standalone trader app connected back into nullclaw over MCP.

PR #487, opened 2026-03-13
fix(websocket): stop TLS connections from closing prematurely on empty reads

0xkuze:
### Summary

v2026.3.12 introduced plain-socket (ws://) support in src/websocket.zig for the OneBot channel. The refactor added a blanket n == 0 → ConnectionClosed check in readExact that applied to both TLS and plain-socket paths.

This broke all wss:// channels (Discord, Lark, DingTalk, QQ, etc.): the TLS readVec can return 0 bytes transiently — e.g. when processing session tickets, re-keying, or empty TLS records — without the connection actually being closed. The check caused these transient zero-returns to be misinterpreted as EOF, triggering an immediate error.ConnectionClosed and putting the gateway into a reconnect loop.

### Root Cause
```zig
// Before (v2026.3.12 — broken):
const n = if (self.tls) |tls| blk: {
    // ... TLS readVec ...
} else self.stream.read(buf[total..]) catch |e| return e;
if (n == 0) return error.ConnectionClosed;  // ← applied to BOTH paths
```
The TLS path already handles real connection closure via error.EndOfStream. The zero-return check is only meaningful for plain sockets (POSIX EOF convention).

### Fix
Confine the bytes == 0 EOF check to the plain-socket branch only. TLS continues to rely on error.EndOfStream as it did in v2026.3.11.

```zig
// After (fixed):
const n = if (self.tls) |tls| blk: {
    // ... TLS readVec (EndOfStream → ConnectionClosed) ...
} else blk: {
    const bytes = self.stream.read(buf[total..]) catch |e| return e;
    if (bytes == 0) return error.ConnectionClosed;  // ← plain-socket only
    break :blk bytes;
};
```

### Validation
``zig build test --summary`` all — 5584/5588 passed, 4 skipped, 0 leaks
``zig build -Doptimize=ReleaseSmall`` — builds clean
``zig fmt --check src/websocket.zig`` — no issues

### Capture Error:
SCR-20260313-bxbi

### Capture Fix:
SCR-20260313-bxnz

### DIFF
--- a/src/websocket.zig
+++ b/src/websocket.zig
@@ -894,10 +894,51 @@ test "ws buildFrame zero-len payload close" {
 }
 
 // Regression: v2026.3.12 applied a blanket `n == 0 → ConnectionClosed` check
-// to both TLS and plain socket paths.  TLS readVec can return 0 transiently
-// (empty TLS records, session tickets, re-keying) without the connection being
-// closed — the old code simply looped.  The fix confines the EOF-by-zero check
-// to the plain-socket branch only.  These tests lock that contract.
+// to both TLS and plain socket paths. TLS readVec may return 0 while it
+// refills its internal buffer or processes post-handshake records, so only
+// plain sockets should treat a zero-byte read as EOF.
+
+fn fake_tls_test_stream(_: *std.Io.Reader, _: *std.Io.Writer, _: std.Io.Limit) std.Io.Reader.StreamError!usize {
+    return error.EndOfStream;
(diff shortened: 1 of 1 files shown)


Nauxscript:
I really need this one. After finally getting all the other Lark channel connection issues sorted out, this is the last piece of the puzzle for me. It's been 6 days since I first installed nullclaw and I still haven't been able to get it working properly 😭

PR #275, opened 2026-03-04
feat: per-agent workspace isolation

DNature:
## Summary

Adds per-agent workspace isolation so named agents can have their own identity files and namespaced memory. When a named agent in `agents.list` specifies `workspace_path`, it gets:

- **Isolated workspace directory** with its own AGENTS.md, SOUL.md, IDENTITY.md, MEMORY.md (auto-scaffolded on first use)
- **Namespaced memory** — scoped to `session_id = "agent:{name}"` so agents can store the same key independently without collisions
- **Cross-agent read-only access** — agents can recall other agents' memories but can only write to their own namespace

### Config example

```json
{
  "agents": {
    "list": [
      {
        "name": "nix",
        "provider": "anthropic",
        "model": "claude-haiku-4-5",
        "workspace_path": "agents/nix"
      }
    ]
  }
}
```

`workspace_path` resolves relative to `NULLCLAW_HOME` (~/.nullclaw/). Absolute paths also supported. Agents without `workspace_path` are fully backward compatible.

### What changed

- **Config** (`config_types.zig`, `config_parse.zig`, `config.zig`): `workspace_path` field on `NamedAgentConfig`, JSON parsing, `resolveAgentWorkspace()` path resolution, `scaffoldAgentWorkspace()` auto-creates directory with starter identity files
- **Memory VTable** (`memory/root.zig`, 10 engine files): Extended `get()` and `forget()` with optional `session_id` parameter for namespace isolation
- **SQLite** (`memory/engines/sqlite.zig`): Idempotent migration from `UNIQUE(key)` to composite `UNIQUE(key, COALESCE(session_id, '__global__'))`, session-scoped get/forget filtering, updated UPSERT conflict clause
- **Agent init** (`session.zig`, `agent/cli.zig`, `subagent.zig`): Workspace resolution wired into session manager (via agent routing), CLI (`--agent` flag), and subagent spawner
- **Memory tools** (`tools/memory_forget.zig`, `memory_store.zig`, `memory_recall.zig`, `memory_list.zig`, `tools/root.zig`): All 4 memory tools bind the agent's `session_id` with per-execution binding for gateway multi-session mode

### Stats

- 35 files changed, +2502 / -370 lines
- 39 new tests (4825 total, 0 leaks)
- Release build: 2.7MB (down from 3.3MB on main)
- Full backward compatibility — no breaking changes

## Test plan

- [x] `zig build test --summary all` — 4825/4826 passed, 1 skipped, 0 leaks
[...]

PR #500, opened 2026-03-13
fix(session): resolve dangling provider pointer in named-agent sessions

vedmalex:
## Summary

Fixes a post-merge regression from PR #450 (topic agent bindings) where named-agent sessions created via `/bind <agent>` would silently fail — the agent received no responses and `/new` would hang.

## Root Cause

In `getOrCreate()`, `Agent.fromConfigWithProfile()` received a `Provider` vtable interface whose `ptr` pointed into a **stack-local** `SessionProviderContext.holder`. When the holder was later moved into `session.provider_holder` via struct assignment, the agent's provider `ptr` became **dangling**.

**Before (broken flow):**
1. `resolveProviderForSession()` → `provider_ctx` with holder on stack
2. `holder.provider()` → vtable `ptr` points into `provider_ctx.holder`
3. `Agent.fromConfigWithProfile(provider_ctx.provider, ...)` → agent stores dangling ptr
4. `session.provider_holder = provider_ctx.holder` → holder moved, agent ptr is stale

**After (fixed flow):**
1. `resolveProviderForSession()` → `provider_ctx` with holder on stack
2. `session.provider_holder = provider_ctx.holder` → move holder to session-owned memory **first**
3. `session.provider_holder.provider()` → vtable `ptr` points into session-owned memory
4. `Agent.fromConfigWithProfile(session_provider, ...)` → agent stores stable ptr

## Live Reproduction

- **Before fix**: `/bind repro` → messages hung with ⚡ reaction, no bot response
- **After fix**: `/bind repro` → bot replied correctly; `/new` → fresh greeting

## Changes

- `src/session.zig`: Reorder provider initialization in `getOrCreate()` so `ProviderHolder` is moved to session-owned memory before `holder.provider()` is called
- Adds regression test: `"getOrCreate stores named agent provider interface from session-owned holder"` — verifies `agent.provider.ptr` points into session-owned `provider_holder`

## Validation

- All tests pass: **5586 passed, 11 skipped, 0 failures, 0 leaks**
- Live Telegram test confirmed fix works end-to-end

Relates to #450

### DIFF
--- a/src/session.zig
+++ b/src/session.zig
@@ -195,20 +195,37 @@ pub const SessionManager = struct {
         errdefer if (!key_owned_by_session) self.allocator.free(owned_key);
 
         const session = try self.allocator.create(Session);
+        session.provider_holder = null;
+        session.owned_provider_api_key = null;
         var session_initialized = false;
         errdefer {
             if (session_initialized) session.deinit(self.allocator);
             self.allocator.destroy(session);
         }
+        errdefer if (!session_initialized) {
+            if (session.provider_holder) |*holder| holder.deinit();
+            if (session.owned_provider_api_key) |key| self.allocator.free(key);
+        };
 
         const agent_profile = findProfileForSessionKey(self.config, session_key);
         var provider_ctx = try self.resolveProviderForSession(agent_profile);
         errdefer provider_ctx.deinit(self.allocator);
 
+        session.* = undefined;
+        session.provider_holder = provider_ctx.holder;
+        session.owned_provider_api_key = provider_ctx.owned_api_key;
+        provider_ctx.holder = null;
+        provider_ctx.owned_api_key = null;
+
+        const session_provider = if (session.provider_holder) |*holder|
+            holder.provider()
+        else
+            provider_ctx.provider;
+
         var agent = try Agent.fromConfigWithProfile(
             self.allocator,
             self.config,
-            provider_ctx.provider,
+            session_provider,
             self.tools,
             self.mem,
             self.observer,
@@ -224,10 +241,12 @@ pub const SessionManager = struct {
             agent.usage_record_ctx = @ptrCast(self);
         }
 
+        const session_provider_holder = session.provider_holder;
+        const session_owned_provider_api_key = session.owned_provider_api_key;
         session.* = .{
             .agent = agent,
-            .provider_holder = provider_ctx.holder,
-            .owned_provider_api_key = provider_ctx.owned_api_key,
+            .provider_holder = session_provider_holder,
+            .owned_provider_api_key = session_owned_provider_api_key,
             .created_at = std.time.timestamp(),
             .last_active = std.time.timestamp(),
             .last_consolidated = 0,
@@ -238,8 +257,6 @@ pub const SessionManager = struct {
         };
         key_owned_by_session = true;
         session_initialized = true;
-        provider_ctx.holder = null;
-        provider_ctx.owned_api_key = null;
 
         // Restore persisted conversation history from session store
         if (self.session_store) |store| {
@@ -1360,6 +1377,30 @@ test "getOrCreate applies named agent profile from routed session key" {
     try testing.expectEqualStrings("ollama", session.agent.default_provider);
     try testing.expectApproxEqAbs(@as(f64, 0.25), session.agent.temperature, 0.000001);
     try testing.expectEqual(@as(usize, 0), session.agent.model_routes.len);
+}
+
+test "getOrCreate stores named agent provider interface from session-owned holder" {
+    var mock = MockProvider{ .response = "ok" };
+    var cfg = testConfig();
+    cfg.default_provider = "openrouter";
+    cfg.agents = &.{
+        .{
+            .name = "Coder Agent",
+            .provider = "ollama",
+            .model = "qwen2.5-coder:14b",
+        },
+    };
+
+    var sm = testSessionManager(testing.allocator, &mock, &cfg);
+    defer sm.deinit();
+
+    const session = try sm.getOrCreate("agent:coder-agent:telegram:group:-100123");
+    try testing.expect(session.provider_holder != null);
+
+    var holder = &session.provider_holder.?;
+    const holder_provider = holder.provider();
+    try testing.expect(session.agent.provider.ptr == holder_provider.ptr);
+    try testing.expect(session.agent.provider.vtable == holder_provider.vtable);
 }
 
 test "getOrCreate falls back to default config for unknown routed agent id" {


ats-bcon:
I retested the fix. I now get "**Mar 13 20:40:26 ubuntu nullclaw[83883]: error(channel_loop): Agent error: error.ApiError**". NullClaw does not crash, but I dont quite know whats wrong. Is it maybe my config?

grafik

ats-bcon:
@vedmalex ,  I found out why the error above accured. 

I would like to clarify that and ask you some questions. If you do not mind, I will do this in this thread. 

I am not sure, if I am finding bugs or if what I see is expected behaviour...

---

1) **The errors like these came from wrong config.json settings.** I have placed _"account_id": "main"_ under "peer" object. This created these errors. Just an idea. Maybe it makes sence to catch misconfiguration and write something a more clear message. (ISSUE SOLVED) 

 ```
Mar 14 12:10:29 ubuntu nullclaw[117371]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:direct:5555555555
Mar 14 12:10:29 ubuntu nullclaw[117371]: **error(channel_loop): Agent error: error.ApiError**
Mar 14 12:10:56 ubuntu nullclaw[117371]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:group:-1009999999999:thread:2
Mar 14 12:10:56 ubuntu nullclaw[117371]: **error(channel_loop): Agent error: error.ApiError**
Mar 14 12:11:55 ubuntu nullclaw[117371]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:group:-1009999999999:thread:4
Mar 14 12:11:56 ubuntu nullclaw[117371]: error(channel_loop): Agent error: error.ApiError
Mar 14 12:12:03 ubuntu nullclaw[117371]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:group:-1009999999999:thread:53
Mar 14 12:12:03 ubuntu nullclaw[117371]: error(channel_loop): Agent error: error.ApiError
Mar 14 12:12:12 ubuntu nullclaw[117371]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:group:-1009999999999
Mar 14 12:12:12 ubuntu nullclaw[117371]: error(channel_loop): Agent error: error.ApiError
Mar 14 12:12:20 ubuntu nullclaw[117371]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:direct:5555555555
Mar 14 12:12:20 ubuntu nullclaw[117371]: error(channel_loop): Agent error: error.ApiError
 ```


2) **Does it change bot behaviour or create a problem if I will NOT add "account_id" at all?** Like following example:

 ```                
 "agent_id": "orchestrator",
                        "match": {
                                "channel": "telegram",
                                "peer": {
                                        "kind": "direct",
                                        "id": "5555555555"
                                }
                        }
 ```

3) **Looking at the log at (1), why does the the session always show "orchestrator"?** In my config, I have bindings to different telegram topics. Also, wenn I trigger /topics in the topic-chat Nr4 (coder) , the bot always claims to be orchestrator on ollama. (It should be coder on openrouter)

**agents**

 ```
        "agents": {
                "defaults": {
                        "model": {
                                "primary": "ollama/kimi-k2.5:cloud"
                        }
                },
                "list": [
                        {
                                "id": "orchestrator",
                                "provider": "ollama",
                                "model": "kimi-k2.5:cloud",
                                "temperature": 0.4,
                                "system_prompt": "You are the Orchestrator. Your job is to manage tasks and delegate them."
                        },
                        {
                                "id": "coder",
                                "provider": "openrouter",
                                "model": "openrouter/moonshotai/kimi-k2.5",
                                "temperature": 0.2,
                                "system_prompt": "You are an expert software engineer. Focus on clean, efficient code."
                        },
                        {
                                "id": "translator",
                                "provider": "ollama",
                                "model": "ollama/kimi-k2.5:cloud",
                                "temperature": 0.6,
                                "system_prompt": "You are a professional translator. Provide accurate and natural translations."
                        }
                ]
        }
 ```

**bindings**

 ```
       "bindings": [
                {
                        "agent_id": "orchestrator",
                        "match": {
                                "channel": "telegram",
                                "account_id": "main",
                                "peer": {
                                        "kind": "direct",
                                        "id": "5555555555"
                                }
                        }
                },
                {
                        "agent_id": "coder",
                        "match": {
                                "channel": "telegram",
                                "account_id": "main",
                                "peer": {
                                        "kind": "group",
                                        "id": "-1009999999999#topic:4"
                                }
                        }
                },
                {
                        "agent_id": "translator",
                        "match": {
                                "channel": "telegram",
                                "peer": {
                                        "kind": "group",
                                        "id": "-1009999999999#topic:53"
                                }
                        }
                }
        ],

```


Telegram topics in topic 4:

```
Telegram topic/session map for chat -1009999999999
Current topic: 4
Active in-memory sessions: 5
• general [idle] turns=1 last_active=296s session=agent:orchestrator:telegram:group:-1009999999999
• topic 2 [idle] turns=1 last_active=302s session=agent:orchestrator:telegram:group:-1009999999999:thread:2
• topic 4 current [idle] turns=1 last_active=317s session=agent:orchestrator:telegram:group:-1009999999999:thread:4
• topic 6 [idle] turns=1 last_active=290s session=agent:orchestrator:telegram:group:-1009999999999:thread:6
• topic 53 [idle] turns=3 last_active=82s session=agent:orchestrator:telegram:group:-1009999999999:thread:53
```

**Comment by vedmalex** [comment]
@ats-bcon great debugging! Let me answer your questions:

---

### 1) `account_id` placement — yes, that's the issue

The config schema expects `account_id` at the **match level**, not inside `peer`:

```json
// ✅ Correct
{
  "agent_id": "coder",
  "match": {
    "channel": "telegram",
    "account_id": "main",  // <-- here
    "peer": {
      "kind": "group",
      "id": "-1009999999999#topic:4"
    }
  }
}

// ❌ Wrong (what you had)
{
  "match": {
    "peer": {
      "account_id": "main"  // <-- inside peer = ignored
    }
  }
}
```

Agreed on better validation — right now unknown fields in `peer` are silently ignored. A config validator warning would help.

---

### 2) Omitting `account_id` — works but less specific

Yes, you can omit `account_id`:

```json
{
  "agent_id": "translator",
  "match": {
    "channel": "telegram",
    "peer": { "kind": "group", "id": "-1009999999999#topic:53" }
  }
}
```

This matches **any** Telegram account configured in nullclaw. If you have only one account, it's fine. If you have multiple, prefer explicit `account_id` to avoid surprises.

---

### 3) Why do ALL sessions show "agent:orchestrator"? 🚨

**This is a bug.** Looking at your logs:

```
• topic 4 [idle] session=agent:orchestrator:telegram:group:-1009999999999:thread:4
```

But your binding says topic 4 → coder. The session key should be:
```
agent:coder:telegram:group:-1009999999999:thread:4
```

**Root cause:** This is exactly the dangling provider pointer bug fixed in this PR (#500). When `getOrCreate()` creates a named-agent session, the provider reference becomes dangling, and the agent routing falls back to the default agent (orchestrator).

**To fix:** 
1. Make sure you're running a build that includes this PR's fix
2. Or apply the patch manually if testing on main
3. After the fix, you'll see correct session keys like `agent:coder:...` and `/topics` will show the right agent

The fix reorders initialization so `ProviderHolder` moves to session-owned memory **before** `holder.provider()` is called — giving stable pointers.

Want me to double-check your build includes this fix?

ats-bcon:
@vedmalex , I think there must still be a bug. I say this, because I compiled the newest code from "main". Look here: 

grafik


grafik



```
Mar 14 13:43:48 ubuntu nullclaw[120638]: Gateway pairing code: 877866
Mar 14 13:43:48 ubuntu nullclaw[120638]: info(memory): memory plan resolved: backend=sqlite retrieval=keyword vector=none rollout=off hygiene=true snapshot=false cache=false semantic_cache=false summarizer=true sources=1
Mar 14 13:43:48 ubuntu nullclaw[120638]: info(channel_manager): telegram polling thread started
Mar 14 13:43:53 ubuntu bash[120639]: {"ok":true,"result":{"message_id":1499,"from":{"id":444444444,"is_bot":true,"first_name":"Nully","username":"nxxxxxx"},"chat":{"id":55555555,"first_name":"xxxxxx","type":"private"},"date":1773492233,"text":"\ud83d\ude80 NullClaw Gateway has been started or restarted !"}}
Mar 14 13:43:53 ubuntu systemd[836]: Started nullclaw.service - nullclaw gateway runtime.
Mar 14 13:45:13 ubuntu nullclaw[120638]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:group:-1009999999999:thread:53
Mar 14 13:45:15 ubuntu ollama[731]: [GIN] 2026/03/14 - 13:45:15 | 200 |   1.80006476s |       127.0.0.1 | POST     "/api/chat"
Mar 14 13:45:52 ubuntu nullclaw[120638]: debug(rollout): rollout decide: mode=off decision=keyword_only session=agent:orchestrator:telegram:group:-1009999999999:thread:4
Mar 14 13:45:54 ubuntu ollama[731]: [GIN] 2026/03/14 - 13:45:54 | 200 |  1.424695249s |       127.0.0.1 | POST     "/api/chat"
Mar 14 13:45:56 ubuntu ollama[731]: [GIN] 2026/03/14 - 13:45:56 | 200 |   1.81771395s |       127.0.0.1 | POST     "/api/chat"
```

P.S. There is a bug in /status , it does not show the version properly.

grafik

vedmalex:
@ats-bcon thanks for the follow-up — you helped us find a **second bug**!

### What's happening

Your session keys all show `agent:orchestrator` because your config uses the `#topic:` format:
```json
"id": "-1009999999999#topic:4"
```

But the route resolver internally constructs peer IDs as:
```
-1009999999999:thread:4
```

Since the comparison is exact string match, your bindings **never match**, and everything falls back to the default agent (orchestrator).

### Why `#topic:` seems natural

You probably copied this format from nullclaw's logs — that's where `#topic:` appears, because `telegramChatTargetAlloc` in `gateway.zig` uses it as the internal wire format for Telegram targets. It's a totally reasonable assumption that the config should use the same format.

### Two ways to fix right now

**Option A** — change your config to use `:thread:` format (the documented format):
```json
{ "peer": { "kind": "group", "id": "-1009999999999:thread:4" } }
{ "peer": { "kind": "group", "id": "-1009999999999:thread:53" } }
```

**Option B** — use `/bind coder` directly in topic 4, `/bind translator` in topic 53. The `/bind` command already saves in the correct format.

### The permanent fix

PR #529 adds automatic normalization: `#topic:` → `:thread:` at config parse time. After that lands, both formats will work in `config.json`.

So you have **two separate bugs** affecting your setup:
1. **#529** — `#topic:` format mismatch (causes wrong agent routing) ← **this is your current issue**
2. **#500** — dangling provider pointer (causes crash/hang after correct routing) ← **you'll hit this next once routing works**

Both need to land for full topic binding support.

ats-bcon:
@vedmalex . Thanks again. I knew there must have been a configuration error on my side. 

_I still think the is an issue or maybe another configuration error._

When I use CONFIG_A, the routing does not work. When I use CONFIG_B. 

The routing works!

grafik


**CONFIG A**  ("account_id": "main", is configured) -> config does not work.

```
        "bindings": [
                {
                        "agent_id": "orchestrator",
                        "match": {
                                "channel": "telegram",
                                "account_id": "main",
                                "peer": {
                                        "kind": "direct",
                                        "id": "555555555"
                                }
                        }
                },
                {
                        "agent_id": "coder",
                        "match": {
                                "channel": "telegram",
                                "account_id": "main",
                                "peer": {
                                        "kind": "group",
                                        "id": "-1009999999999:thread:4"
                                }
                        }
                },
                {
                        "agent_id": "translator",
                        "match": {
                                "channel": "telegram",
                                "account_id": "main",
                                "peer": {
                                        "kind": "group",
                                        "id": "-1009999999999:thread:53"
                                }
                        }
                }
        ],
```

**CONFIG B: "account_id": "main" is NOT configured. config WORKS!**


 ```
       "bindings": [
                {
                        "agent_id": "orchestrator",
                        "match": {
                                "channel": "telegram",
                                "peer": {
                                        "kind": "direct",
                                        "id": "555555555"
                                }
                        }
                },
                {
                        "agent_id": "coder",
                        "match": {
                                "channel": "telegram",
                                "peer": {
                                        "kind": "group",
                                        "id": "-1009999999999:thread:4"
                                }
                        }
                },
                {
                        "agent_id": "translator",
                        "match": {
                                "channel": "telegram",
                                "peer": {
                                        "kind": "group",
                                        "id": "-1009999999999:thread:53"
                                }
                        }
                }
        ],

```

-----

P.S. If you find time, would you fix the bug that prevents showing nullclaw version in /status? It is so conenient to copy paste the information when creating bug reports :-D

**Comment by vedmalex** [comment]
@ats-bcon mystery solved! 🔍 It's **not a code bug** — it's a configuration mismatch.

## The Root Cause

Your Telegram config doesn't specify `account_id`, so it defaults to `"default"`. But your binding expects `"main"`:

**Your binding (CONFIG A):**
```json
"account_id": "main"
```

**Your Telegram config (likely):**
```json
{
  "channels": {
    "telegram": {
      "bot_token": "..."
      // No account_id here! Defaults to "default"
    }
  }
}
```

**The mismatch:**
```
Binding expects: "main"
Runtime has:    "default"
Result:         NO MATCH → fallback to default agent
```

## Why CONFIG B works

When you **omit** `account_id` from the binding:
```json
// No account_id field
"match": { "channel": "telegram", "peer": {...} }
```

The binding says "match ANY account_id", so it passes the pre-filter and successfully matches on the peer (Tier 1).

## The Fix

**Option 1** — Add `account_id` to your Telegram config (recommended):
```json
{
  "channels": {
    "telegram": {
      "account_id": "main",
      "bot_token": "YOUR_BOT_TOKEN",
      ...
    }
  }
}
```

**Option 2** — Remove `account_id` from bindings (match any account):
```json
{
  "agent_id": "coder",
  "match": {
    "channel": "telegram",
    // Remove this: "account_id": "main",
    "peer": { "kind": "group", "id": "-1009999999999:thread:4" }
  }
}
```

## Technical Details

The matching happens in `agent_routing.zig:219-227`:
```zig
if (binding.match.account_id) |ba| {
    if (!std.mem.eql(u8, ba, input.account_id)) return false;
}
```

Your binding has `account_id: "main"`, so it enters this branch and compares against the runtime's `"default"`. They don't match, so the binding is excluded from candidates before peer matching even happens.

---

**This is working as designed** — the account_id filtering prevents cross-account binding leakage. The confusion comes from the implicit `"default"` value when not specified in the Telegram config.

Hope this finally clears it up!

ats-bcon:
Ah, got it. This was an easy one. I didnt see the account_id attribute in "channels" object. This thread will be great for people who might get confused by this in the future.

ats-bcon:
@vedmalex . This pull might have introduced a small issue : https://github.com/nullclaw/nullclaw/pull/275 by @DNature 

It is now possible to setup a specific agent-directory like this: 

grafik

In that case the "system_prompt" (which can be a text or a file) is omitted. 

grafik

I am not sure, if it is a problem or not, but I am very sure this will create confusion as soon as people will start working with those attributes. At first my expectation was that even when I setup a dedicacted directory for the sub-agent it should still read the prompt and know its name (which he does not). 

So maybe the best solution here is to explicitly document which attributes get ommited in case when a "workspace_directory" is specified. Or maybe its best to remove "system_prompt" attribute, now that we have the option for dedicated agent directory.

vedmalex:
Follow-up on the latest `workspace_path` / `system_prompt` question:

I checked this again on current `main` and could not reproduce a runtime regression. On the current code path, `workspace_path` and `system_prompt` are applied together:
- named-agent routed sessions keep the profile prompt
- `/subagents spawn --agent <id>` keeps the profile prompt
- the dedicated workspace still provides its own bootstrap context

To make this explicit and prevent future confusion, I pushed `de45d32` to this PR branch with:
- regression test for routed/named-agent sessions
- regression test for subagent task-runner path
- docs clarification in `README.md`, `docs/en/configuration.md`, and `docs/zh/configuration.md`

Validation on this branch:
- `zig build test --summary all` -> `5857 passed`, `11 skipped`, `0 failures`

So my current conclusion is: this looks like a documentation/discoverability issue rather than a `#500` behavior regression.

vedmalex:
Follow-up for this discussion is now split into a separate PR: #620

Link: https://github.com/nullclaw/nullclaw/pull/620

That PR contains only the regression tests and documentation clarification for `workspace_path` + `system_prompt` behavior.

ats-bcon:
Very cool. Thanks. I just assumed that "system_prompt" gets ommited, because my bot did not know its name, even though the name is stated in the prompt. I further assume, that the "system_prompt" gets overruled by the first boot up prompt, which forces the model to say  that it just came online and has no clue who it is.

ats-bcon:
P.S. For those, who care. 

I just tested out what wins. The name in the "system_prompt" or the name in the IDENTITY.md. The IDENTITY.md wins. The bot always responds with the name in the IDENTITY.md

grafik

PR #586, opened 2026-03-16
fix(a2a): log underlying error on 'Agent processing failed'

vernonstinebaker:
## Summary

- Both the streaming and non-streaming A2A handlers used a bare `catch { }` block that silently discarded the original error before returning the opaque `"Agent processing failed"` JSON-RPC response. Operators had no way to diagnose the real cause from logs.
- Add a scoped logger (`std.log.scoped(.a2a)`) and capture the error name at both catch sites so the root error appears in the gateway log alongside the task ID.

## Validation

```
zig build test --summary all
# 5719/5723 passed, 4 skipped (pre-existing), 0 leaks
```

`zig fmt --check src/a2a.zig` — clean.

No unit test added: this is a logging-only / pure error-propagation change. A meaningful test would require a mock `SessionManager`. An inline comment documents this decision per `AGENTS.md §8.1`.

## Notes

- No behavioral change for callers — the JSON-RPC error response is identical. Only the gateway log gains new information.
- The scoped logger tag `.a2a` follows the pattern used by other modules in the codebase.

### DIFF
--- a/src/a2a.zig
+++ b/src/a2a.zig
@@ -13,6 +13,8 @@ const gateway = @import("gateway.zig");
 const ConversationContext = @import("agent/prompt.zig").ConversationContext;
 const streaming = @import("streaming.zig");
 
+const log = std.log.scoped(.a2a);
+
 /// Maximum number of tasks kept in the registry before eviction.
 const MAX_TASKS: usize = 1000;
 const A2A_PROTOCOL_VERSION = "0.2.5";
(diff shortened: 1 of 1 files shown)


vernonstinebaker:

vernonstinebaker:
Closing confirmed — this fix landed independently via commit `10b01b7` ("fix(a2a): log failed turns for debugging") by @dprusne on 2026-03-20, which applies the exact same changes: scoped `log` declaration and `catch |err|` with `log.err()` at both A2A handler sites. No action needed.

PR #594, opened 2026-03-16
fix(providers): handle reasoning_content in streaming for reasoning models (glm-5, kimi-k2.5)

### LINKED ISSUE #576: moonshot-intl provider fails with kimi-k2.5
`nullclaw agent -m "hi" --provider moonshot-intl --model kimi-k2.5` returns `error: ApiError`.

Direct curl to `https://api.moonshot.ai/v1/chat/completions` with the same API key and model returns a valid response. The API works, nullclaw does not parse it.

kimi-k2.5 returns a `reasoning_content` field alongside `content` in its responses.

Steps to reproduce:

```json
{
  "models": {
    "providers": {
      "moonshot-intl": {"api_key": "sk-..."}
    }
  },
  "agents": {
    "defaults": {
      "model": {"primary": "moonshot-intl/kimi-k2.5"}
    }
  }
}
```

```bash
nullclaw agent -m "hi"
```

Version: dev (commit 26a5f55, main branch)
OS: Linux

vernonstinebaker:
## Summary

- **Bug**: Reasoning models (glm-5, kimi-k2.5, minimax-m2.x) that emit thinking via `delta.reasoning_content` with `"content": ""` caused `NoResponseContent` errors in streaming mode, surfacing as "Agent processing failed" to clients.
- **Root cause**: `extractDeltaContent` in `sse.zig` discarded empty `content` strings and never checked `reasoning_content`, so the entire thinking phase was silently skipped. The accumulated stream buffer stayed empty → `stream_result.content = null` → agent saw empty `response_text` → `NoResponseContent`.
- **Fix (`sse.zig`)**: `extractDeltaContent` now falls back to `delta.reasoning_content` when `content` is absent or empty. Reasoning chunks are wrapped in `<think>...</think>` so `ThinkStripStreamCallback` suppresses them from the visible stream; users only see the final answer.
- **Fix (`compatible.zig`)**: Both the `CurlWaitError`/`CurlFailed` fallback path and the large-request non-streaming skip path now check `reasoning_content` when `content` is null, strip think-blocks before emitting, and correctly free all `ChatResponse` fields (`tool_calls`, `reasoning_content`) that were previously not freed in the fallback paths.

Fixes #576

## Validation

```
zig fmt --check src/          ✅ clean
zig build test --summary all  ✅ 5727/5731 passed, 4 skipped, 1 leaked (pre-existing: gateway.test.run returns AddressInUse)
zig build -Doptimize=ReleaseSmall  ✅ clean
```

Baseline before this change: 5723/5727 passed — our 4 new regression tests account for the difference. The pre-existing gateway leak is unchanged.

End-to-end tested against a live gateway running glm-5 (infini-ai): a vikunja-mcp tool-calling query that previously always failed with `NoResponseContent` now streams successfully to completion.

## Notes

- No config changes, no new public API — purely internal streaming parser behavior.
- The `<think>` wrapping approach reuses the existing `ThinkStripStreamCallback` state machine already in `compatible.zig`; no new abstraction needed.
- Risks: none identified. The fallback only activates when `content` is empty/absent, so models that populate `content` normally are unaffected.

### DIFF
--- a/src/providers/compatible.zig
+++ b/src/providers/compatible.zig
@@ -793,12 +793,41 @@ pub const OpenAiCompatibleProvider = struct {
                 .{ self.name, request_text_bytes, self.max_streaming_prompt_bytes.? },
             );
             const fallback = try chatImpl(ptr, allocator, request, model, temperature);
-            if (fallback.content) |text| {
-                callback(callback_ctx, root.StreamChunk.textDelta(text));
+            // Emit visible content. For reasoning models (e.g. glm-5, kimi-k2.5)
+            // `content` may be null while `reasoning_content` holds the answer;
+            // strip think-blocks and emit whichever is non-empty.
+            const emit_text: ?[]const u8 = if (fallback.content) |c| c else fallback.reasoning_content;
+            if (emit_text) |text| {
+                const stripped = try stripThinkBlocks(allocator, text);
+                defer allocator.free(stripped);
+                if (stripped.len > 0)
+                    callback(callback_ctx, root.StreamChunk.textDelta(stripped));
             }
             callback(callback_ctx, root.StreamChunk.finalChunk());
+            // Build the returned content: prefer stripped content, fall back to stripped reasoning.
+            const returned_content: ?[]const u8 = blk: {
+                if (fallback.content) |c| {
+                    const s = try stripThinkBlocks(allocator, c);
+                    if (s.len > 0) break :blk s;
+                    allocator.free(s);
+                }
+                if (fallback.reasoning_content) |rc| {
+                    const s = try stripThinkBlocks(allocator, rc);
+                    if (s.len > 0) break :blk s;
+                    allocator.free(s);
+                }
+                break :blk null;
+            };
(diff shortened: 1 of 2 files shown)


vernonstinebaker:

vernonstinebaker:
Closing as superseded. The two core fixes in this PR landed upstream independently via other contributors:

- **`sse.zig` — `extractDeltaContent` fallback to `delta.reasoning_content`** (wrapping in `<think>…</think>`): landed in PR #578 (`manelsen/feat/issue-576-moonshot-kimi-response`), commit `73b45dc`.
- **`compatible.zig` — `reasoning_content` handling in non-streaming path and null-content guard**: landed in PR #641 (`yanggf8/fix/glm-thinking-and-tool-calls`), commit `61b2310`.

Both implementations independently arrived at the same approach. All other file changes bundled in this PR (`a2a.zig`, `agent/root.zig`, `config.zig`, `config_parse.zig`, etc.) landed through their own respective PRs (#584, #590, #591, etc.).

PR #608, opened 2026-03-17
fix(providers): fall back to non-streaming on CurlWaitError for…

brunojuliao:
fall back to non-streaming on CurlWaitError for Anthropic, Gemini, and Vertex

The compatible (OpenAI) provider already catches CurlWaitError/CurlFailed and gracefully falls back to a non-streaming chatImpl() call. However, the Anthropic, Gemini, and Vertex providers propagated these errors directly, causing gateway failures on resource-constrained platforms (e.g. aarch64/Raspberry Pi) where the async child.wait() in the streaming curl path returns error.SystemResources.

This aligns all three providers with the existing fallback pattern in compatible.zig, so streaming failures degrade to non-streaming requests instead of crashing the agent loop.

### DIFF
--- a/src/providers/openai_codex.zig
+++ b/src/providers/openai_codex.zig
@@ -452,6 +452,7 @@ fn codexStreamRequest(
     var saw_text_delta = false;
     var emitted_text_fallback = false;
     var emitted_tool_payload = false;
+    var saw_terminal = false;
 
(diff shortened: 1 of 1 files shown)


brunojuliao:
That looks much better! Thank you, @DonPrus 😇

PR #537, opened 2026-03-14
fix(security): enforce secrets encryption / 修复(security): 强制执行机密加密

manelsen:
## Summary
This PR enforces secrets encryption by making `secrets.encrypt=true` a mandatory requirement for all configurations. It prevents the application from running with insecure plaintext secret storage.

### Key Changes
- **Configuration Validation**: Added `InsecurePlaintextSecrets` to `Config.ValidationError`. The `validate()` method now rejects any configuration where `secrets.encrypt` is set to `false`.
- **Onboarding Guard**: Updated the onboarding wizard (`onboard.zig`) and all setup modes (interactive, quick, channels-only) to verify that encryption is enabled before proceeding.
- **Improved Error Messaging**: Provides clear guidance when a configuration is rejected due to disabled encryption.
- **Security Baseline**: Establishes a higher security floor by ensuring that sensitive tokens and keys are always stored in an encrypted format.

## Validation
n## Notes
- Closes #523.
- `zig build test --summary all`: Added new test cases:
    - `validation rejects plaintext secrets` in `src/config.zig`.
    - `ensureSecretsEncryptionEnabled rejects plaintext secrets config` in `src/onboard.zig`.
- Verified that the onboarding wizard correctly blocks setup if a user manually disables encryption in an existing config.

---

### 摘要
本 PR 强制执行机密加密，将 `secrets.encrypt=true` 设置为所有配置的强制要求。它防止应用程序在不安全的明文机密存储状态下运行。

### 主要改动
- **配置验证**: 在 `Config.ValidationError` 中增加了 `InsecurePlaintextSecrets`。`validate()` 方法现在会拒绝任何将 `secrets.encrypt` 设置为 `false` 的配置。
- **引导程序守卫**: 更新了引导向导 (`onboard.zig`) 和所有设置模式（交互式、快速设置、仅限渠道），在继续操作之前验证加密功能已启用。
- **改进错误提示**: 当配置因禁用加密而被拒绝时，提供清晰的指导。
- **安全基准**: 通过确保敏感令牌和密钥始终以加密格式存储，建立了更高的安全底线。

### 验证
- `zig build test --summary all`: 新增了测试用例：
    - `src/config.zig` 中的 `validation rejects plaintext secrets`。
    - `src/onboard.zig` 中的 `ensureSecretsEncryptionEnabled rejects plaintext secrets config`。
- 验证了如果用户在现有配置中手动禁用加密，引导向导会正确拦截设置流程。

### 备注
- 关联并关闭 #523。

PR #609, opened 2026-03-17
fix(config): encrypt persisted api keys in config.json

manelsen:
## Summary
- encrypt persisted `api_key` values when `Config.save()` writes `config.json`
- decrypt those `api_key` values during config parsing so runtime behavior stays unchanged
- add regression coverage for encrypted save/load round-trips across providers, reliability fallbacks, model routes, named agents, memory API/Qdrant, and Composio

## Why
A reviewer on #537 correctly pointed out that `secrets.encrypt=true` only enforced plaintext-secret rejection and Nostr key encryption, while regular `api_key` entries were still written to `config.json` in plaintext.

## Validation

Refs: https://github.com/nullclaw/nullclaw/pull/537#issuecomment-4075513735 (thanks again, [ats-bcon](https://github.com/ats-bcon))

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -1,6 +1,7 @@
 const std = @import("std");
 const platform = @import("platform.zig");
 const provider_names = @import("provider_names.zig");
+const secrets = @import("security/secrets.zig");
 pub const config_types = @import("config_types.zig");
 pub const config_parse = @import("config_parse.zig");
 /// Write a JSON-escaped string (with enclosing quotes) to any writer.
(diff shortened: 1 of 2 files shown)


DonPrus:
One residual note, but it is not about this PR: `BrowserComputerUseConfig.api_key` exists in the types, but the `browser.computer_use` sub-config is not currently serialized or parsed on `main` at all, so this PR neither breaks nor fixes that.

ats-bcon:
@manelsen , If I understand it correctly the following keys are getting encrypted. 
 
- models.providers.*.api_key
- model_routes.*.api_key
- agents.*.api_key
- reliability.api_keys[]
- memory.search.store.qdrant_api_key                                                                                                                                                                                                        - memory.api.api_key
- composio.api_key


I suggest to also encrypt the _bot_token_ and maybe some other secrets in the channels config.
- **channels.telegram.accounts.default.bot_token**

PR #620, opened 2026-03-18
docs: clarify workspace_path and system_prompt behavior

vedmalex:
## Summary

Follow-up to the discussion in #500 about whether `workspace_path` suppresses a named agent's `system_prompt`.

This PR does not change the runtime behavior. It makes the current behavior explicit and adds regression coverage for it.

## What changed

- add a regression test proving routed/named-agent sessions keep the profile `system_prompt` when `workspace_path` is set
- add a regression test proving `/subagents spawn --agent <id>` keeps the profile `system_prompt` when `workspace_path` is set
- document that `workspace_path` and `system_prompt` are combined, not mutually exclusive

## Why

The current runtime already applies both:
- dedicated workspace bootstrap context comes from the agent workspace
- the named profile prompt is still prepended to the system prompt

That behavior was not stated clearly enough in the docs, which made the `#500` thread look like a possible regression.

## Validation

- [x] `zig build test --summary all` -> `5857 passed`, `11 skipped`, `0 failures`

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,6 +525,7 @@ Named agent profiles are configured separately from bindings. Bindings only choo
 
 If a named agent should run from its own workspace, set `agents.list[].workspace_path`.
 Relative paths are resolved from the directory that contains `config.json`, the workspace is scaffolded on first use, and the agent gets a durable memory namespace `agent:<agent-id>`.
+Setting `workspace_path` does not disable `system_prompt`: when both are configured, the named profile prompt is still applied and the workspace bootstrap files are loaded from that dedicated workspace.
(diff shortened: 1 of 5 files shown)


ats-bcon:
P.S. Anybody reading this and wondering why his system_prompt path might NOT be loading. The system path has to be absolute. If the path is not absolute (as in my following example), the agent treats the entire string as the literal system prompt. This means the AI would receive the text of the path itself as its instructions, rather than the contents of the file.

{
                                "id": "lawyer",
                                "provider": "ollama",
                                "model": "minimax-m2.5:cloud",
                                "workspace_path": "workspace/agents/lawyer",
                                "temperature": 0.4,
                                "system_prompt": "~/.nullclaw/workspace/agents/translator/systemprompt-real-estate-lawyer-agent.md"
                        }

PR #711, opened 2026-03-23
Feat/cross memory

DonPrus:
## Summary

This PR adds a deterministic memory event stream to `nullclaw` so memory can be synchronized across agent instances without changing the normal single-agent workflow.

In practice, this gives `nullclaw` the primitives needed for cases like:

- Agent B knows your preferences because Agent A learned them
- multiple agents share or merge behavioral memory
- `nullboiler` can orchestrate memory sync externally

This does **not** make agents manage each other automatically. It only adds the feed/apply/checkpoint primitives needed for external orchestration.

## Why

Before this PR, memory semantics differed too much by backend for deterministic shared memory to be reliable:

- some backends collapsed keys without `session_id`
- some scoped delete paths were not exact
- some recall paths lost scope
- there was no consistent event feed / replay contract
- backend choice changed observable memory behavior

That was acceptable for local storage, but not for deterministic multi-agent memory sync.

### Unified backend semantics

Backends were aligned around the same exact scoped memory model:

- memory identity is effectively `(instance_id, key, session_id)`
- scoped CRUD behaves consistently
- scoped recall respects `session_id`
- `get`, `getScoped`, `list`, `delete`, `deleteScoped`, and `count` no longer drift by backend

### Deterministic event feed

Memory backends now expose a real sync contract:

- `events`
- `apply`
- `status`
- `checkpoint`
- `compact`

This turns memory synchronization into an explicit deterministic feed/replay model instead of ad-hoc state copying.

### Typed behavioral merge operations

The event model now supports more than plain CRUD replication. It includes deterministic merge operations for shared behavioral memory:

- `merge_object`
- `merge_string_set`

This is meant for preferences / traits style memory where overwrite-only semantics are not enough.

### Feed lifecycle for production use

The feed now supports:

- checkpoint export / restore
- compaction
- cursor floor
- `CursorExpired` handling
- sequence continuity after compaction / restore

That makes long-lived synchronization practical instead of requiring unbounded replay forever.

### Native feed support across first-class backends

Native feed implementations were added for:

- `sqlite`
- `postgres`
- `redis`
- `api`
- `clickhouse`
- `markdown`
- `lancedb`
- `memory`

### HTTP sync surface

The API backend / gateway now exposes explicit sync routes:

- `GET /memory/events`
- `POST /memory/apply`
- `GET /memory/status`
- `POST /memory/compact`
- `GET /memory/checkpoint`

### Vector-aware runtime wiring

`lancedb` now receives the runtime embedding provider correctly instead of being initialized with `null`, so vector-aware behavior works in real runtime wiring instead of only in isolated tests.

## Design Intent

The intended model is:

- `nullclaw` owns deterministic memory feed/apply primitives
- `nullclaw` does **not** automatically synchronize agents by itself
- orchestration belongs outside, e.g. in `nullboiler`
- single-agent usage should remain normal and familiar
- shared-memory is opt-in via feed/apply, not forced into the default loop

That is similar in spirit to how `a2a` is integrated: the capability exists in `nullclaw`, but orchestration is external.

## Testing

This PR adds coverage for:

- backend contract tests for exact scoped memory behavior
- native feed roundtrip tests
- `events -> apply` convergence tests
- compaction / checkpoint / restore tests
- cursor-floor behavior
- route-level API / gateway tests
- minimal cross-backend sync coverage
- behavioral merge sync coverage
- runtime wiring regression tests, including `lancedb` embedder wiring

Current full test run on this branch is green.

## Notes

- Out-of-band writes directly to backend storage are **not** part of the deterministic sync contract.
- Intermediate compatibility paths introduced during branch development were intentionally pruned where they only added complexity.
- This PR is large because it introduces a real sync substrate, not just a CLI convenience.

## Outcome

After this PR, `nullclaw` has the primitives needed for deterministic shared memory across agents:

- export events
- apply events
- compact and checkpoint feeds
- restore lagging peers
- deterministically merge some memory types
- keep backend semantics consistent enough that backend choice does not change the core sync model


DonPrus:
Instead of pushing the full cross-memory design directly into nullclaw in [nullclaw/nullclaw#711](https://github.com/nullclaw/nullclaw/pull/711), I moved that responsibility into a separate module/service: [nullclaw/nullpantry](https://github.com/nullclaw/nullpantry).

That PR made it clear that this is not just a small memory feature. It needs deterministic event replay, feed/apply/checkpoint/compact, exact scoped memory semantics across backends, merge operations for behavioral memory, and a sync surface that is closer to multi-primary replication in database terms.

By moving this layer into nullpantry, the same foundation can support the original cross-agent memory use case and broader scenarios like shared/team memory, governed storage, retrieval, lifecycle, vector/graph projections, and production backends.

This keeps nullclaw small, fast, and focused on local execution and baseline memory, while still making it highly extensible through a clear external memory boundary.

PR #716, opened 2026-03-25
Add calculator tool for mathematical operations

festoinc:
## Summary
- Adds `src/tools/calculator.zig` — a single `calculator` tool with 20 operations: `add`, `subtract`, `multiply`, `divide`, `pow`, `sqrt`, `log`, `ln`, `exp`, `average`, `median`, `stdev`, `min`, `max`, `count`, `percentile`, `abs`, `floor`, `ceil`, `round`
- Registered in `src/tools/root.zig` factory (included in `allTools`)

## Rationale
LLMs are natively unreliable at arithmetic — even simple calculations like addition or division frequently produce wrong results, and more complex operations like stdev or percentile are essentially guaranteed to fail. The model often consumes many tokens reasoning through a calculation only to arrive at an incorrect answer. A native calculator tool eliminates this class of hallucination entirely: the LLM delegates the math, gets a precise result in one tool call, and uses the output directly in its response. This reduces both token waste and factual errors.

## Validation
- `zig build test --summary all` — 6127 passed, 11 skipped (pre-existing), 0 failures, 0 leaks
- `zig fmt --check src/tools/calculator.zig src/tools/root.zig` — clean

## Notes
- No docs changes needed (tool is internal, not user-facing CLI)
- 30 unit tests covering all operations and error paths
- Population stdev (divide by N), percentile with linear interpolation
- Max 6 decimal places output, trailing zeros stripped
- Explicit error handling for invalid inputs (division by zero, negative sqrt, non-positive log, wrong arg count, out-of-range percentile)

### DIFF
--- a/src/tools/calculator.zig
+++ b/src/tools/calculator.zig
@@ -11,9 +11,9 @@ pub const CalculatorTool = struct {
     const MIN_FIXED_NOTATION_ABS: f64 = 0.000001;
 
     pub const tool_name = "calculator";
-    pub const tool_description = "Perform mathematical calculations accurately. Supports arithmetic (add, subtract, multiply, divide, pow, sqrt), logarithms (log, ln, exp), rounding (abs, floor, ceil, round), and statistics (average, median, stdev, min, max, count, percentile).";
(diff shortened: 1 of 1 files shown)


montvid:
Now the llm seems to use python to calculate - is this pr somehow better? Any other agents that have calculator tool?

festoinc:
@montvid yes. Zeroclaw does. Using python is possible only if python is allowed by security policy, so python is not that bulletproof

DonPrus (approved):

ats-bcon:
Seems to work. Pretty cool!

grafik

PR #719, opened 2026-03-26
fix(session): correct Discord scheduled DM delivery target | 修复(session): 更正 Discord 定时私聊投递目标

### LINKED ISSUE #715: Discord scheduled DM delivery may use an invalid target identifier
Summary:
Scheduled Discord DM delivery may be reusing a direct-session identifier as the outbound delivery target, even though Discord message send requires a valid DM channel target.

Observed behavior:
- normal live Discord DM replies work
- delayed or scheduled DM delivery is suspicious and may not resolve to a valid send target

Reason for concern:
- scheduled delivery appears to persist and reuse the delivery target field
- Discord outbound send ultimately targets the channels/{channel_id}/messages REST path
- if the stored direct-session identifier is not an actual DM channel target, delayed delivery can fail even while live DM replies still work

Minimal repro:
1. Configure one Discord bot account and route it to a normal agent.
2. Start nullclaw and open a DM with that bot.
3. Send a normal message first so the direct session exists.
4. Ask the agent to create a one-shot reminder for about 1 minute later.
5. Wait for the scheduled delivery.

Expected:
- the reminder is delivered back into the same Discord DM

Actual:
- normal live replies can work, but scheduled DM delivery may not resolve to a valid Discord send target

Relevant code areas:
- src/channel_adapters.zig
- src/session.zig
- src/tools/schedule.zig
- src/cron.zig
- src/channels/discord.zig

Possible fixes:
- reject unsupported Discord DM scheduled delivery cases explicitly, or
- add a Discord-specific resolution step from direct-session identity to a valid DM channel target before sending.

manelsen:
### EN:
   - Fixed a bug where scheduled Discord DM delivery could reuse the direct-session user identifier instead of the real DM channel target.
   - Added `delivery_chat_id` to normalized conversation context so delayed delivery can preserve the correct outbound target without changing peer-based session routing.
   - Updated session schedule context wiring to prefer the preserved delivery target while keeping routing metadata keyed by peer identity.
   - Added regression coverage for Discord scheduled DM delivery and inbound conversation-context propagation.

### ZH:
   - 修复了 Discord 定时私聊投递可能复用 direct session 用户标识、而不是实际 DM channel 目标的问题。
   - 在标准化 conversation context 中新增 `delivery_chat_id`，让延迟投递能够保留正确的 outbound target，同时不改变基于 peer 的 session 路由。
   - 更新了 session 的 schedule context 传递逻辑，优先使用保留下来的 delivery target，同时继续保持基于 peer identity 的路由元数据。
   - 为 Discord 定时私聊投递以及入站 conversation context 传递增加了回归测试。

## Validation
   - zig build test --summary all

## Notes
   - Risk: Medium. The change touches session scheduling context, but does not alter Discord session-key derivation or general peer routing semantics.
   - Fixes #715

### DIFF
--- a/src/channel_loop.zig
+++ b/src/channel_loop.zig
@@ -895,6 +895,7 @@ fn processTelegramMessage(
     const conversation_context = buildConversationContext(.{
         .channel = "telegram",
         .account_id = tg_ptr.account_id,
+        .delivery_chat_id = sender,
         .peer_id = sender,
         .is_group = is_group,
         .group_id = if (is_group) sender else null,
@@ -1742,6 +1743,7 @@ pub fn runSignalLoop(
                 .account_id = sg_ptr.account_id,
                 .sender_number = if (msg.sender.len > 0 and msg.sender[0] == '+') msg.sender else null,
                 .sender_uuid = msg.sender_uuid,
+                .delivery_chat_id = msg.reply_target orelse msg.sender,
                 .peer_id = if (msg.is_group) msg.group_id else msg.sender,
                 .group_id = msg.group_id,
                 .is_group = msg.is_group,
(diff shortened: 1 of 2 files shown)


babutree:
Thanks, this looks like the right fix and the added regression coverage is very helpful.

PR #744, opened 2026-03-29
fix(ollama): normalize scheduler tool aliases | 修复(ollama): 规范 scheduler 工具别名

manelsen:
### EN:
   - Normalized `scheduler_tool` and `schedule_tool` to the canonical `schedule` tool name in the Ollama tool-call adapter.
   - Kept the change scoped to Ollama's existing tool-name quirk handling, alongside the current `tool.` and `tools.` prefix normalization.
   - Added regression coverage for both alias extraction and the formatted tool-call response path so the scheduler tool remains reachable when a local model emits the alias spelling.
   - Fixes #743.

### ZH:
   - 在 Ollama 的 tool-call 适配层中，将 `scheduler_tool` 和 `schedule_tool` 规范化为标准工具名 `schedule`。
   - 变更范围严格限定在 Ollama 现有的工具名兼容处理逻辑内，与当前 `tool.` / `tools.` 前缀归一化保持同一层级。
   - 补充了回归测试，覆盖 alias 提取以及格式化后的 tool-call 响应路径，确保本地模型输出该别名时仍能正确调用 scheduler 工具。
   - 修复 #743。

## Validation
   - `zig fmt --check src/providers/ollama.zig`

## Notes
   - Risk: Low. The change is limited to Ollama tool-name normalization before the agent dispatcher sees the call.
   - No docs change needed; the canonical runtime tool name remains `schedule`.
   - `zig test src/providers/ollama.zig` is not a valid standalone validation path in this repository layout, so validation was run through the normal build entrypoint instead.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -81,6 +81,12 @@ Defined in `src/root.zig`. Phases mirror deployment dependencies:
 - `src/security/` - Policy enforcement (`policy.zig`), pairing (`pairing.zig`), encrypted secrets (`secrets.zig`), sandbox backends (`landlock.zig`, `firejail.zig`, `bubblewrap.zig`, `docker.zig`, `detect.zig`).
 - `src/agent/` - Agent loop internals: `dispatcher.zig` (tool call parsing), `compaction.zig` (history trimming), `prompt.zig` (system prompt builder), `memory_loader.zig` (context injection), `commands.zig` (agent-mode commands). Config defaults are `max_tool_iterations = 1000` and `max_history_messages = 100` (see `src/config_types.zig`).
 
+### Provider Boundary Notes
+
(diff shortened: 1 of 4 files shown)


ats-bcon:
Great, but I wonder, why this fix was necessary. IMHO this worked some releases ago with ollama.

manelsen:
I haven't traced back the commit that broke the feature. I have fixed the code in a way tests show, and I believe, that will work.

Complexity sometimes brings regression. Adding new features comes at the price of sometimes losing what we took for safe and done.

We have fences and tests to minimize the chance of regression; alas, it still happens.

I've been doing some deduplication, renewing the wiring, looking for better ways to do things on the backend. Still, some PRs await review and there's only so much I can do without merging or refusing what's already done.

DonPrus (approved):

ats-bcon:
@manelsen . I dont think the fix works.

grafik

grafik

PR #770, opened 2026-04-05
feat(api): REST Admin API — runtime status, config read, model listing, and cron job management

vernonstinebaker:
## Summary

Adds an opt-in REST Admin API under `/api/` that gives lightweight clients (menubar apps, iOS/iPadOS, CLI dashboards) a uniform, structured interface for gateway inspection and control.

- **Zero new dependencies.** Single static Zig binary; estimated binary size increase < 30 KB.
- **Opt-in.** Disabled by default (`gateway.admin_api: false`). Returns `403 ADMIN_API_DISABLED` until explicitly enabled.
- **Auth mirrors `/cron`.** Reuses `extractBearerToken` + `isWebhookAuthorized` from `gateway.zig`. No new auth surface.
- **Standard JSON envelope everywhere.** `{"success":true,"data":{...},"error":null}` / `{"success":false,"data":null,"error":{"code":"...","message":"..."}}`.
- **`src/api/` abstraction layer.** Handlers live in `api.zig`; `gateway.zig` calls `api.dispatch()`. No circular imports.

## Endpoints

```
GET    /api/status               version, pid, uptime, overall_status, components
GET    /api/config?path=<dotted> read a config value (on-disk, dotted path)
GET    /api/models               provider list — names + has_key, never key values
GET    /api/cron                 list all scheduled jobs
POST   /api/cron                 create recurring cron job
POST   /api/cron/once            create one-shot delayed job
POST   /api/cron/:id/run         trigger immediately
POST   /api/cron/:id/pause       pause job
POST   /api/cron/:id/resume      resume job
PATCH  /api/cron/:id             update job fields
DELETE /api/cron/:id             remove job
```

`/health` is unchanged (plain `{"status":"ok"}` for external health checkers).
`/api/cron/*` wins over legacy `/cron` handlers — the `/api/` routing block is first in the chain.

## Key design decisions

- **No version prefix.** `/api/v1/` was rejected (YAGNI — no v2 caller); flat `/api/` prefix chosen.
- **`/api/status` subsumes `/api/health`.** Single endpoint returns version + pid + uptime + overall status + component map; no separate health endpoint.
- **`/api/` routing block at the top** of the `gateway.zig` if/else chain so structured cron endpoints take precedence over the legacy `/cron` handlers.
- **Scheduler passed pre-locked.** `gateway.zig` calls `lockSharedSchedulerForRequest()` before `api.dispatch()` and unlocks after return; handlers never touch the mutex.

## Files changed

| File | Change |
|------|--------|
| `src/api/api.zig` | New — registry, dispatcher, 11 handlers, path-param matching, 26 tests |
| `src/api/context.zig` | New — per-request context struct, `sendSuccess`/`sendError` helpers |
| `src/gateway.zig` | Wire `/api/` prefix check at top of routing chain; pass locked scheduler to `api.dispatch()` |
| `src/config_types.zig` | Add `GatewayConfig.admin_api: bool = false` |
| `src/config_parse.zig` | Parse `gateway.admin_api` from JSON |
| `config.example.json` | Document `admin_api` field |
| `docs/en/gateway-api.md` | Full endpoint table, auth, envelope, and `GET /api/status` response shape |

## Test results

```
6312/6318 passing, 0 leaks
```

Pre-existing failure: `agent.cli.test.writeRateLimitHint mentions reliability knobs and logs` — crashes on `main` as well, unrelated to this PR.

## Validation

```bash
zig build test --summary all   # 6312/6318 passing, 0 leaks
zig fmt --check src/            # clean
```

### DIFF
--- a/src/api/api.zig
+++ b/src/api/api.zig
@@ -1,0 +1,304 @@
+//! REST Admin API — v1 router and endpoint registry.
+//!
+//! All endpoints live under /api/v1/.  The surface is opt-in: when
+//! `gateway.admin_api` is false (the default) every request to /api/v1/*
+//! receives a 403 with a clear error message so clients can surface a
+//! useful hint rather than a silent 404.
+//!
+//! Adding a new endpoint in a later phase is a one-step operation:
+//!   1. Write a handler fn(ctx: *ApiContext) anyerror!void in this file
+//!      (or import it from a dedicated file for larger phases).
+//!   2. Append an Endpoint entry to the `endpoints` slice in `registry`.
+//!
+//! Auth model (mirrors the /cron pattern from gateway.zig):
+//!   - No pairing guard configured → allow (gateway running without auth).
+//!   - Pairing disabled → allow.
+//!   - Pairing enabled, no tokens yet → deny (bootstrap phase).
+//!   - Pairing enabled, tokens present → require valid Bearer token.
+//!
+//! Path parameter convention: for paths with a dynamic segment (e.g.
+//! /api/v1/models/:name) the Endpoint.path field ends with "/:param" and
+//! matchPath() returns the segment after the last "/" as the path param.
+
+const std = @import("std");
+const builtin = @import("builtin");
+const health = @import("../health.zig");
+const Config = @import("../config.zig").Config;
+const ApiContext = @import("context.zig").ApiContext;
+
+// ── Endpoint registry ────────────────────────────────────────────────
+
+pub const Endpoint = struct {
+    /// HTTP method string, e.g. "GET".
+    method: []const u8,
+    /// URL path, e.g. "/api/v1/health".
+    /// A trailing "/:param" suffix indicates a single dynamic path segment.
+    path: []const u8,
+    /// Handler function.  Must not panic; errors are caught by the dispatcher.
+    handler: *const fn (ctx: *ApiContext) anyerror!void,
+};
+
+/// Comptime-built slice of all registered /api/v1/* endpoints.
+/// Phase 0 registers only /api/v1/health.  Later phases append more.
+pub const registry: []const Endpoint = &.{
+    .{ .method = "GET", .path = "/api/v1/health", .handler = handleHealth },
+};
+
+// ── Dispatcher ───────────────────────────────────────────────────────
+
+/// Result returned to gateway.zig so it can write the wire response.
+pub const DispatchResult = struct {
+    /// HTTP status line, e.g. "200 OK".
+    status: []const u8,
+    /// Response body.
+    body: []const u8,
+    /// True when body was heap-allocated and must be freed with `allocator`.
+    allocated: bool,
+};
+
+/// Main entry point called by gateway.zig for every request whose base_path
+/// starts with "/api/v1/".
+///
+/// `pairing_guard_ptr` is `?*const PairingGuard` erased to `?*anyopaque` so
+/// this file does not need to import the pairing module directly; the gateway
+/// passes the already-checked boolean `auth_ok` instead.
+///
+/// Parameters:
+///   allocator    — request-scoped arena allocator
+///   raw_request  — full raw HTTP bytes
+///   method       — HTTP method string
+///   target       — full request target (may include query string)
+///   base_path    — target without query string
+///   config_opt   — active config or null
+///   auth_ok      — true when the bearer token has already been validated
+///                  by the caller (gateway.zig) using isWebhookAuthorized.
+pub fn dispatch(
+    allocator: std.mem.Allocator,
+    raw_request: []const u8,
+    method: []const u8,
+    target: []const u8,
+    base_path: []const u8,
+    config_opt: ?*const Config,
+    auth_ok: bool,
+) DispatchResult {
+    // Guard: admin_api must be explicitly enabled in gateway config.
+    const enabled = if (config_opt) |cfg| cfg.gateway.admin_api else false;
+    if (!enabled) {
+        return .{
+            .status = "403 Forbidden",
+            .body = "{\"success\":false,\"data\":null,\"error\":{\"code\":\"ADMIN_API_DISABLED\",\"message\":\"Set gateway.admin_api=true in config.json to enable the REST admin API\"}}",
+            .allocated = false,
+        };
+    }
+
+    // Guard: bearer auth.
+    if (!auth_ok) {
+        return .{
+            .status = "401 Unauthorized",
+            .body = "{\"success\":false,\"data\":null,\"error\":{\"code\":\"UNAUTHORIZED\",\"message\":\"Valid Bearer token required\"}}",
+            .allocated = false,
+        };
+    }
+
+    // Route: walk registry for exact path + method match, then try prefix for
+    // dynamic segments.
+    var ctx = ApiContext{
+        .allocator = allocator,
+        .raw_request = raw_request,
+        .method = method,
+        .target = target,
+        .base_path = base_path,
+        .config_opt = config_opt,
+    };
+
+    for (registry) |ep| {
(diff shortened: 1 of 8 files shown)


src/api/api.zig
```diff
@@ -0,0 +1,1590 @@
...
+///   MISSING_PARAM  — path query parameter not provided
+///   CONFIG_ERROR   — path not found or config unreadable
+fn handleConfig(ctx: *ApiContext) anyerror!void {
+    // Extract path query param from target, e.g. /api/config?path=foo.bar
+    const path_param = extractQueryParam(ctx.target, "path") orelse {
+        try ctx.sendError("400 Bad Request", "MISSING_PARAM", "query parameter 'path' is required");
+        return;
+    };
+
+    const value_json = config_mutator.getPathValueJson(ctx.allocator, path_param) catch |err| {
```
manelsen, line 336: No path filtering here before the read. `getPathValueJson` has no allowlist — it will happily return `models.providers.0.api_key` or any other credential stored in `config.json`. Needs a deny-list for sensitive suffixes (`api_key`, `token`, `secret`, `pairing`) or a read-specific allowlist before this is safe to ship.

manelsen (changes requested):
First off — this is genuinely impressive work. The architecture is clean, the `src/api/` separation is well thought out, the JSON envelope is consistent throughout, auth gating is correctly placed before any dispatch, `GET /api/models` deliberately returns `has_key` instead of the key value, the mutex handling mirrors the existing `/cron` pattern exactly, and the test suite is thorough. This is the kind of PR that makes a codebase better just by existing.

I have one blocker that needs a small fix before I can recommend merge, and a few minor notes below.

---

## Blocker — `GET /api/config` can expose API keys

`handleConfig` calls `config_mutator.getPathValueJson(ctx.allocator, path_param)` with the path coming directly from the query string, with no filtering. `isAllowedPath` in `config_mutator` is only enforced on *mutations*, not reads. So an authenticated client who knows the config schema can do:

```
GET /api/config?path=models.providers.0.api_key
→ {"success":true,"data":{"path":"...","value":"sk-live-..."}}
```

Same for pairing tokens, webhook secrets, or any other credential stored in `config.json`.

The PR description says the intent is to read "non-sensitive values like `default_provider`" — the implementation just doesn't enforce that yet.

**Suggested fix** — one of these, whichever feels right to you:

Option A — explicit deny-list in `handleConfig` (fast, surgical):
```zig
const sensitive_suffixes = [_][]const u8{ "api_key", "token", "secret", "password", "pairing" };
for (sensitive_suffixes) |suffix| {
    if (std.mem.endsWith(u8, path_param, suffix)) {
        try ctx.sendError("403 Forbidden", "PATH_FORBIDDEN", "that config path is not readable via the API");
        return;
    }
}
```

Option B — allowlist (safer long-term): reuse or extend `isAllowedPath` from `config_mutator` to cover readable paths too.

---

## Minor notes (non-blocking)

**`sendError` message safety** — the `message` parameter is interpolated directly into a JSON string via `{s}`. Today every call site passes a string literal or `@errorName(err)` (which is always ASCII-safe), so there's no practical risk. But a doc comment noting "callers must ensure `message` contains no JSON-special characters" would prevent a future regression.

**`GET /api/config` path injection** — you already call `jsonEscapeString` on `path_param` when embedding it in the response (line 339), which is correct. Just noting it's good.

**Test coverage** — I looked at all 78 tests. Coverage of auth, routing, envelope shape, and error paths is solid. One gap worth considering: a test asserting that `GET /api/config?path=models.providers.0.api_key` returns 403 after the fix lands, to lock in the regression protection.

---

Really looking forward to seeing this merged once the config read path is tightened up — the Admin API layer is going to be very useful. Happy to re-review quickly once updated. @DonPrus tagging you for visibility.

vernonstinebaker:
Thanks for the thorough review, @manelsen — the config read path analysis is spot on. Here is our plan before we push a revision:

## Blocker — `GET /api/config` path filtering

We will go with **Option B (allowlist)**, matching the approach used for mutations.

**Step 1 — Add `pub fn isAllowedReadPath` to `src/config_mutator.zig`**

A new, separate allowlist for reads — narrower than the mutation list because read access has a different threat model. The mutation allowlist allows `channels.`, `security.`, and `models.providers.` prefixes that can all contain API keys and tokens. The read allowlist will cover only clearly non-sensitive dotted paths (e.g. `default_temperature`, `reasoning_effort`, `gateway.port`, `gateway.host`, `tunnel.provider`, `agents.defaults.model.primary`, and the `memory.` and `scheduler.` prefixes). The `channels.*`, `security.*`, and `models.providers.*` prefixes will be absent from the read list.

**Step 2 — Guard `handleConfig` in `src/api/api.zig`**

After extracting `path_param`, add before the `getPathValueJson` call:

```zig
if (!config_mutator.isAllowedReadPath(path_param)) {
    try ctx.sendError("403 Forbidden", "PATH_FORBIDDEN", "that config path is not readable via the API");
    return;
}
```

**Step 3 — Add regression tests**

Two new test cases in `src/api/api.zig`:
- `GET /api/config?path=models.providers.0.api_key` → 403 `PATH_FORBIDDEN`
- `GET /api/config?path=channels.telegram.accounts.default.bot_token` → 403 `PATH_FORBIDDEN`

Plus a positive case (`default_temperature`) to confirm the allowlist is wired correctly.

Also add tests for `isAllowedReadPath` in `config_mutator.zig` covering at least one forbidden prefix and one allowed path.

---

## Minor notes we will also address

**`sendError` doc comment** — we will add a `// NOTE: message must contain no JSON-special characters (callers pass string literals or @errorName output only)` comment above the function in `context.zig`.

**Docs markdown typo** — the diff introduces a leading space on `## Endpoints` (becomes ` ## Endpoints`). We will restore the correct heading while keeping the line change.

---

Once these are in we will run `zig build test --summary all` to confirm 0 failures and 0 leaks before pushing the revision.

vernonstinebaker:
All review comments have been addressed — summary below.

**Blocker — `GET /api/config` key exposure (Option B, allowlist)**
Added `pub fn isAllowedReadPath(path: []const u8) bool` to `config_mutator.zig` with an explicit allowlist of 11 safe paths (`default_provider`, `default_model`, `gateway.port`, `gateway.host`, `gateway.admin_api`, `gateway.rate_limit`, `gateway.timeout_secs`, `gateway.max_body_bytes`, `scheduler.enabled`, `scheduler.max_jobs`) plus prefix matches for `diagnostics.` and `scheduler.`. `handleConfig` returns `403 PATH_FORBIDDEN` for anything not on the list. Security-sensitive paths (`models.providers.*.api_key`, `channels.*.bot_token`, `security.*`, etc.) are absent by design. Tests added for both the forbidden-path 403 and a passing allowed-path case.

**`sendError` message safety**
Added a doc comment to `sendError` in `context.zig` noting that callers must ensure `message` contains no JSON-special characters.

**Docs heading typo**
Fixed the leading-space on the `## Endpoints` heading in `gateway-api.md`.

**Zig 0.16 compatibility**
The branch has been rebased onto `main` (`362e93f`) and `api.zig` updated to the Zig 0.16 API: `std.Io.Writer.Allocating` writer pattern, `health.snapshot(allocator)` with `defer deinit`, and `[]SnapshotComponent` slice iteration. Builds and tests clean (1 pre-existing failure in `tools.web_search` also present on `main`).

DonPrus:
This is a really good idea. That said, I think the better place for the HTTP server is `nullhub`, since it already has one.

In this model, the data flow would be:

`nullclaw -> nullhub -> HTTP`

I believe most of the data needed across these four PRs is already available, but I’ll double-check it next week. If anything is missing, I’ll add it.

Alternatively, you could open a PR in `nullhub`:
https://github.com/nullclaw/nullhub

PR #788, opened 2026-04-08
fix(config): validate agent provider

fakhriaunur:
## Summary
- Add missing validation in `Config.validate()` that rejects named agents with unrecognized provider names
- Follows the same pattern as the existing `ReservedMainAgentName` check (commit 38ee80f)
- Adds `ValidationError.UnknownAgentProvider` enum variant and `isKnownProviderName()` helper

## Validation
- `zig build` compiles clean
- `zig fmt --check src/` passes
- Regression test confirms validation rejects unknown provider names
- `zig build -Doptimize=ReleaseSmall` compiles clean

## Notes
- **Root cause**: `Config.validate()` checked for reserved agent names (`main`) but never validated that `agents.list[].provider` maps to a known provider in `self.providers`. An agent configured with a misspelled or nonexistent provider name would be accepted silently.
- **Fix**: Added `isKnownProviderName()` helper that scans `self.providers` for the given name, and a validation check in the `agents` loop that returns `ValidationError.UnknownAgentProvider` if not found.
- **Impact**: Additive-only config validation. Existing valid configs are unaffected. Invalid configs (previously accepted silently) will now fail at startup with a clear error message.
- **Risk**: LOW — follows the exact same pattern as the existing `ReservedMainAgentName` check at line 1348. No vtable changes, no security surface changes.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -179,6 +179,13 @@ fn namedAgentUsesReservedRootId(agent_name: []const u8) bool {
     return std.mem.eql(u8, agent_routing.normalizeId(&agent_buf, agent_name), "main");
 }
 
+fn isKnownProviderName(providers: []const ProviderEntry, name: []const u8) bool {
+    for (providers) |p| {
+        if (std.mem.eql(u8, p.name, name)) return true;
+    }
+    return false;
+}
+
 // ── Top-level Config ────────────────────────────────────────────
 
 pub const Config = struct {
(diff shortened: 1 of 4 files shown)


fakhriaunur:
## Followup: Fixed CI failure — skip provider validation when providers list is empty

### What broke
The initial validation ran unconditionally against `self.providers`, which defaults to `&.{}` in many configs. Tests that load configs without explicit providers (e.g., `persistBindingUpdate`) would hit `ValidationError.UnknownAgentProvider` before any user-facing validation could occur.

### Fix
Added guard: only validate agent provider names when providers are explicitly configured.

```zig
if (self.providers.len > 0 and !isKnownProviderName(self.providers, agent_cfg.provider)) {
    return ValidationError.UnknownAgentProvider;
}
```

This is correct because:
- When `providers` is empty, the provider comes from the factory (auto-loaded), not user config
- When `providers` is non-empty, the user explicitly configured them, so typos should be caught
- Matches the principle: validate user input, not internal defaults

DonPrus (approved):

PR #789, opened 2026-04-08
Fix gateway bind and rate limit safeguards

ScottTPfaff:
## Summary
- enforce gateway public-bind safety unless an active tunnel is present or allow_public_bind is explicitly enabled
- key gateway rate limiting by route plus client identifier so one sender cannot throttle every client on the same route
- JSON-escape saved string arrays and add regression coverage for the gateway and config writer

## Validation
- zig build
- zig build test --summary all


eabase:
### BUMP

I keep getting rate limit errors with my **free** OpenRouter keys.
Something is not right, because same keys works in other tools.

DonPrus (approved):

PR #796, opened 2026-04-10
ci: add Nix flake build workflow

jonathanhfmills:
## Summary

Adds `.github/workflows/nix.yml` — a CI job that builds and smoke-tests the nullclaw Nix flake on every push and PR.

- Runs on `ubuntu-latest` (x86_64-linux) as `CI / Test (nixos-latest)`
- Uses `cachix/install-nix-action` to set up Nix with flakes enabled
- Uses `nix-community/cache-nix-action` to cache the Nix store between runs (keyed on `*.nix` + `flake.lock`), avoiding redundant zig2nix bootstrapping
- Smoke-tests the built binary

**Note on user namespaces:** Ubuntu 24.04 runners restrict unprivileged user namespace creation by default (via AppArmor). Nix's sandbox and zig2nix's bootstrap build both use `bwrap`, which requires them. Two `sysctl` calls re-enable them with `|| true` fallbacks for compatibility across runner versions.

## Test plan

- [x] Verified on `jonathanhfmills/nullclaw` fork (jonathanhfmills/nullclaw#1) — passes consistently including cache hit runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,43 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  nix-build:
+    name: Test (nixos-latest)
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # Ubuntu 24.04 restricts user namespace creation via AppArmor by default.
+      # Nix's sandbox (bwrap) and zig2nix's bootstrap build both need them.
+      - name: Allow user namespaces (required for zig2nix bwrap on Ubuntu 24.04)
+        run: |
+          sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0 || true
+          sudo sysctl -w kernel.unprivileged_userns_clone=1 || true
+
+      - uses: cachix/install-nix-action@v31
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+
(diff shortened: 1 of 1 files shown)


jonathanhfmills:
**Optional next step: public Cachix binary cache**

This CI validates the Nix build on every PR, but end users who run `nix build github:nullclaw/nullclaw` still compile from source each time (including the zig2nix bootstrap). With enough users, that's a lot of redundant CPU time — and energy.

Setting up a [Cachix](https://www.cachix.org/) cache ([free tier, 5 GB, open-source](https://www.cachix.org/pricing)) would let users' Nix daemons transparently download a pre-built binary instead of compiling. Every user who gets a cache hit skips the full zig2nix bootstrap + compile — what takes ~3 minutes of CI time becomes a few-second download. At scale across the Nix community, that's a meaningful reduction in redundant computation and carbon footprint.

**Getting started:** https://docs.cachix.org/getting-started

Once set up, the CI job uploads on every successful main build, and users substitute with:

```nix
nix.settings = {
  substituters = [ "https://nullclaw.cachix.org" ];
  trusted-public-keys = [ "nullclaw.cachix.org-1:<signing-key>" ];
};
```

Requires a Cachix account and the signing key as a GitHub Actions secret — maintainer-only. Happy to help with the workflow addition if you go that route.

jonathanhfmills:
After sleeping on this PR, I've realized the native Ubuntu CI test might be redundant. Since the `nix-build` already uses Ubuntu as a base, we could potentially replace the native test with the new workflow (or a similar variation).

I trust your judgment, @DonPrus. Hope all's well on your end! LMK if you need anything from me :rocket:

PR #815, opened 2026-04-14
fix(web_search): add setup guidance for missing providers | 修复(web_search): 为缺失搜索提供方补充设置指引

### LINKED ISSUE #812: http_request
### Description

I came from Picoclaw, where I also tested ZeroClaw. I liked NullClaw and it seems very organized. But I've been trying to enable the internet search capability since Friday; I've tested all the examples and it doesn't work. I have the latest version of NullClaw installed.

Can anyone help me?

Another question: where does the NullClaw community exchange messages? Why is nobody talking or responding to anything on the Discord channel? Do they use Lark or WeChat? What is it? It would be good to share experiences and improve.

Best regards

### Expected behavior

_No response_

### Steps to reproduce

_No response_

### Version

_No response_

### OS

None

manelsen:
### EN:
   - Improved `web_search` failure reporting when no reliable search provider is actually configured.
   - Added setup guidance that points users to `http_request.search_base_url` for SearXNG and the expected API key env vars for hosted providers.
   - Kept aggregate provider-failure reporting for configured SearXNG deployments and added regression tests for both behaviors.

### ZH:
   - 改进了 `web_search` 在实际未配置可靠搜索提供方时的失败提示。
   - 新增设置指引，明确提示用户通过 `http_request.search_base_url` 配置 SearXNG，或为托管搜索提供方设置所需的 API key 环境变量。
   - 对已配置 SearXNG 的场景保留聚合失败报告，并为两种行为补充了回归测试。

## Validation
   - `zig build test --summary all`: passed in `/tmp/nullclaw-issue-812`.

## User Impact
   - Users who invoke web search without a working provider now get an actionable setup message instead of a generic multi-provider failure blob.
   - Existing configured search deployments keep the detailed provider failure summary they already rely on.

## Notes
   - Closes #812.

### DIFF
--- a/src/tools/web_search.zig
+++ b/src/tools/web_search.zig
@@ -32,6 +32,8 @@ const DEFAULT_COUNT: usize = 5;
 const DEFAULT_TIMEOUT_SECS: u64 = 30;
 /// Upper bound for provider chain size (primary + fallbacks + auto expansions).
 const MAX_PROVIDER_CHAIN: usize = 16;
+const WEB_SEARCH_SETUP_HINT =
+    "web_search is not configured with a reliable provider. Configure http_request.search_base_url for SearXNG or set one of BRAVE_API_KEY, FIRECRAWL_API_KEY, TAVILY_API_KEY, PERPLEXITY_API_KEY, EXA_API_KEY, JINA_API_KEY, or WEB_SEARCH_API_KEY. DuckDuckGo is best-effort only.";
 
 const SearchProvider = enum {
     auto,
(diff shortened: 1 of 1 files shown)


DonPrus (approved):

uMendex:
I've already tested six projects from the claw family, and all of them search the internet without any problems. Only nullclaw is a headache and doesn't work at all, and no developer seems concerned about solving this root problem of the project.

The solution can't be paying tokens to Brave. All the other projects work directly with DuckDuckGo, so why doesn't NullClaw work? Running on a $5 device makes it unpredictable that it will work directly with DuckDuckGo.

PR #872, opened 2026-04-28
fix: Zig 0.16 Mattermost empty-body POST and gateway accept-loop CPU spin

vernonstinebaker:
> [!CAUTION]
> **High-severity regression** — both issues affect all Mattermost-connected agents in production since the Zig 0.16 migration:
> - **100% CPU utilisation** on the gateway thread (busy-spin on EAGAIN) — affects all platforms in daemon mode
> - **Silent Mattermost messaging failure** — all bot messages silently dropped (empty POST body)

## Summary

Two bugs introduced by the Zig 0.16 migration, both observed on live hardware (OrangePi RV2 / riscv64, Radxa Zero 3W / aarch64).

- **Mattermost silent failure**: `std.Io.Writer.Allocating` does not flush to its backing `ArrayList` until `toArrayList()` is called. The previous code passed `body.items` to `curlPost()` before calling `toArrayList()`, so every POST was sent with an empty body and Mattermost silently rejected all messages. Fixed by extracting `buildTypingRequestBody`, `buildPostRequestBody`, and `buildDirectChannelRequestBody` helpers that each call `toArrayList()` before returning an owned slice.

- **Gateway 100% CPU spin**: In daemon mode the listen socket is set non-blocking (`force_nonblocking = daemon_mode`). Under Zig 0.16 `std.Io.Threaded`, `netAcceptPosix` maps `EAGAIN` from `accept4()` to `errnoBug()` instead of `error.WouldBlock`. In non-debug builds `errnoBug` returns `error.Unexpected`, which the accept loop's `else => continue` caught without sleeping — producing a tight spin of ~9,600 `accept4` calls/second and ~100% CPU on the gateway thread. Fixed by adding an explicit `error.Unexpected` arm that sleeps `ACCEPT_POLL_INTERVAL_MS` (100 ms) identically to the `error.WouldBlock` arm. In debug builds the same bug manifests as a panic (`programmer bug caused syscall error: AGAIN`); this fix resolves that too. Also replaced `std_compat.thread.sleep`'s `std.Io.sleep` call with `std.c.nanosleep` (POSIX) / `Sleep()` (Windows), which reliably suspends the calling thread at the OS level.

## Validation

```
zig build test --summary all
→ 6713 pass, 9 skip, 1 fail (6723 total)
```

The 1 failure (`tools.web_search` — aggregate error message check) and the 9 failures seen when run from inside a git hook (`skills.installSkillFromGit` — git subprocess inherits `GIT_DIR` from hook environment) are both pre-existing on the base commit at the same counts. Confirmed by running the test suite on the base commit under both environments.

Regression tests added for all three Mattermost body builders (`buildTypingRequestBody`, `buildPostRequestBody`, `buildDirectChannelRequestBody`).

Also includes a pre-existing `zig fmt` fix for `src/mcp_admin.zig` (required for the pre-commit hook to pass).

## Notes

- The `netAcceptPosix` EAGAIN→`errnoBug` mapping is a Zig 0.16 stdlib bug; this fix is a targeted workaround at the gateway accept loop. A proper fix would require patching `std/Io/Threaded.zig`.
- Verified on live hardware: CPU dropped from ~98% sustained to ~0% idle on both boards after deploying the fix.
- Risk tier: **High** (`gateway.zig` change) + **Medium** (`channels/mattermost.zig`, `compat.zig`).


vernonstinebaker:
Closing in favour of a clean, focused PR. This branch was accidentally based on a commit that included unrelated REST API work, causing ~40 unrelated commits and API source files to appear in the diff. New PR will contain only the two Zig 0.16 regression fixes.

PR #853, opened 2026-04-19
fix(gateway): add accept-loop backoff to prevent CPU spin | 修复(gateway): 为 accept 循环增加退避，避免 CPU 空转

manelsen:
Fixes #851

### EN:
- Hardened the gateway accept loop to apply bounded backoff on non-WouldBlock accept errors, preventing tight CPU spin under repeated transient failures.
- Added explicit backoff helpers/constants (`ACCEPT_ERROR_BACKOFF_MAX_MS`, log interval control) so retry behavior is deterministic and observable.
- Added regression coverage in `src/gateway.zig` for the backoff progression and cap behavior.

### ZH:
- 加固了 gateway 的 accept 循环：在非 WouldBlock 的 accept 错误下应用有上限的退避，避免瞬时错误反复发生时出现 CPU 空转。
- 增加了明确的退避辅助逻辑与常量（`ACCEPT_ERROR_BACKOFF_MAX_MS`、日志节流间隔），使重试行为可预测且可观测。
- 在 `src/gateway.zig` 增加回归测试，覆盖退避增长与上限行为。

## Validation
- `zig build test --summary all`

## Notes
- This change is intentionally scoped to gateway accept-loop error handling and does not alter successful connection flow.

PR #873, opened 2026-04-28
fix: Zig 0.16 Mattermost empty-body POST and gateway accept-loop CPU spin

vernonstinebaker:
> [!CAUTION]
> **High-severity regression** — both issues affect all Mattermost-connected agents in production since the Zig 0.16 migration:
> - **100% CPU utilisation** on the gateway thread (busy-spin on EAGAIN) — affects all platforms in daemon mode
> - **Silent Mattermost messaging failure** — all bot messages silently dropped (empty POST body)

## Summary

Two bugs introduced by the Zig 0.16 migration, both observed on live hardware (OrangePi RV2 / riscv64, Radxa Zero 3W / aarch64).

**Mattermost silent failure** (`src/channels/mattermost.zig`): `std.Io.Writer.Allocating` does not flush to its backing `ArrayList` until `toArrayList()` is called. The previous code passed `body.items` to `curlPost()` before calling `toArrayList()`, so every POST was sent with an empty body and Mattermost silently rejected all messages. Fixed by extracting `buildTypingRequestBody`, `buildPostRequestBody`, and `buildDirectChannelRequestBody` helpers that each call `toArrayList()` before returning an owned slice.

**Gateway 100% CPU spin** (`src/gateway.zig`, `src/compat.zig`): In daemon mode the listen socket is set non-blocking (`force_nonblocking = daemon_mode`). Under Zig 0.16 `std.Io.Threaded`, `netAcceptPosix` maps `EAGAIN` from `accept4()` to `errnoBug()` instead of `error.WouldBlock`. In non-debug builds `errnoBug` returns `error.Unexpected`, which the accept loop's `else => continue` caught without sleeping — producing a tight spin of ~9,600 `accept4` calls/second and ~100% CPU on the gateway thread. Fixed by adding an explicit `error.Unexpected` arm that sleeps `ACCEPT_POLL_INTERVAL_MS` (100 ms) identically to the `error.WouldBlock` arm. Also replaced `std_compat.thread.sleep`'s `std.Io.sleep` call with a direct `std.c.nanosleep` call on POSIX (reliably suspends the OS thread; `std.Io.sleep` does not yield the CPU in the Threaded IO context). Windows retains `std.Io.sleep` via IOCP which works correctly there.

## Commits

- `style: format mcp_admin.zig` — pre-existing fmt issue caught by pre-commit hook
- `fix(channels/mattermost): finalize allocating writer body before curlPost`
- `fix(gateway, compat): CPU spin from EAGAIN on non-blocking accept socket`
- `fix(compat): Windows sleep guard (std.Io.sleep via IOCP on Windows, std.c.nanosleep on POSIX)`

## Validation

```
zig build test --summary all
→ 6603 pass, 9 skip, 1 fail (6613 total)
```

The 1 failure (`tools.web_search` — aggregate error message check) is pre-existing on the base commit. The 9 failures seen inside the git hook environment (`skills.installSkillFromGit` — inherits `GIT_DIR`) are also pre-existing. Regression tests added for all three Mattermost body builders.

Verified on live hardware: CPU dropped from ~98% sustained to ~0% idle on both arm boards after deploying the fix.

## Notes

- The `netAcceptPosix` EAGAIN→`errnoBug` mapping is a Zig 0.16 stdlib issue; this fix is a targeted workaround at the accept loop. A proper fix would require patching `std/Io/Threaded.zig`.
- Risk tier: **High** (`gateway.zig`) + **Medium** (`channels/mattermost.zig`, `compat.zig`).

### DIFF
--- a/src/channels/mattermost.zig
+++ b/src/channels/mattermost.zig
@@ -12,6 +12,71 @@ const Atomic = @import("../portable_atomic.zig").Atomic;
 
 const log = std.log.scoped(.mattermost);
 
+fn buildTypingRequestBody(
+    allocator: std.mem.Allocator,
+    channel_id: []const u8,
+    thread_id: ?[]const u8,
+) ![]u8 {
+    var body: std.ArrayListUnmanaged(u8) = .empty;
+    errdefer body.deinit(allocator);
+    var body_writer: std.Io.Writer.Allocating = .fromArrayList(allocator, &body);
+    const bw = &body_writer.writer;
+    try bw.writeAll("{\"channel_id\":");
+    try root.appendJsonStringW(bw, channel_id);
+    if (thread_id) |tid| {
+        if (tid.len > 0) {
+            try bw.writeAll(",\"parent_id\":");
+            try root.appendJsonStringW(bw, tid);
+        }
+    }
+    try bw.writeByte('}');
+    body = body_writer.toArrayList();
+    return try body.toOwnedSlice(allocator);
+}
+
+fn buildPostRequestBody(
+    allocator: std.mem.Allocator,
+    channel_id: []const u8,
+    text: []const u8,
+    thread_id: ?[]const u8,
+) ![]u8 {
+    var body: std.ArrayListUnmanaged(u8) = .empty;
+    errdefer body.deinit(allocator);
+    var body_writer: std.Io.Writer.Allocating = .fromArrayList(allocator, &body);
+    const bw = &body_writer.writer;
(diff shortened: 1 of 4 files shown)


vernonstinebaker:
Closing in favour of #853 which covers the same three changes (compat.thread.sleep nanosleep fix, gateway accept-loop, Mattermost body builder refactor) and is more complete on the accept loop (exponential backoff + logging vs. just handling error.Unexpected). The regression test comment on buildPostRequestBody may be worth preserving in #853.

Issue #192, opened 2026-03-01
Eliminate external subprocess dependencies (curl and wasmtime)


lambda-alpha-labs:
### Description

This feature replaces the current use of external subprocesses for core functionality with native, in-process implementations. Specifically, it involves refactoring the HTTP networking layer to use Zig's native std.http.Client instead of shelling out to the curl binary, and updating the WASM runtime adapter to statically link a lightweight C-based WASM interpreter (such as wasm3 or the wasmtime C-API) rather than invoking the wasmtime CLI via child processes.

### Motivation

NullClaw's primary value proposition is being a strictly self-contained, lightweight static binary that requires no external dependencies. Relying on host-installed tools like curl and wasmtime breaks this guarantee of true portability and introduces unnecessary performance bottlenecks due to continuous process-spawning overhead. By bringing these capabilities entirely in-process, we ensure the application remains completely standalone, reduces invocation latency, and fully adheres to its zero-dependency design philosophy while preserving a minimal memory footprint.

ThinkOffApp:
This matters a lot for embedded targets. We just shipped ClawWatch — NullClaw running natively on a Samsung Galaxy Watch (ARM, 2 GB RAM). On a watch, there is no curl binary, no wasmtime, no standard unix toolchain. The static binary approach works precisely because NullClaw can be self-contained.

Having zero subprocess dependencies means the binary drops into any ARM Linux environment (Android, embedded Linux, Raspberry Pi, smartwatches) and just runs. This is a real blocker for production embedded deployments.

Our build: `zig build -Dtarget=arm-linux-musleabihf -Doptimize=ReleaseSmall` — produces a 2.8 MB static binary that runs on the watch via ProcessBuilder.

Repo: https://github.com/ThinkOffApp/ClawWatch

manelsen:
I might have figured it out. Gimme a few days.

PR #881, opened 2026-05-01
refactor(http): remove runtime curl subprocesses

ncode:
Summary: replace curl-backed Zig HTTP helpers with native std.http wrappers; migrate providers, channels, gateway, tools, memory API, update, voice, and SSE paths; rename runtime HTTP helpers and errors from Curl* to Http*.

Boundary: keep curl as auxiliary/operator tooling for Docker builds, compose healthchecks, docs/examples, and shell-safety fixtures; the removal is limited to NullClaw runtime subprocess execution.

Validation: zig build test --summary all (6781/6790 passed, 9 skipped); git diff --check.

Risk: broad network migration touches gateway, tools, and security-adjacent paths; SOCKS proxy URLs now fail closed because Zig 0.16 std.http supports HTTP(S) proxies only.

Addresses the curl part from https://github.com/nullclaw/nullclaw/issues/192

Thank you for the awesome project, and I'm very new to Zig. I've really tested most flows related to Telegram, web_search across different providers and OpenAI connections and tried to add as much test coverage as possible. The real downside I see from the patch is the removal of SOCKS5 support (and I've discovered it late, but since I already started the work I thought I'd open the PR anyway).

I've also thought about adding the native provider in parallel, not sure what's the opinion on that.


mark-os:
Is Zig ready for non-curl HTTP? I looked into it and it only supports TLS 1.3? This will break lots of `web_fetch` requests. Maybe have a chance to fall back to system curl?

addadi (changes requested):
## Review: REQUEST CHANGES

High-risk refactor requiring additional work before merge.

### Critical Issues (Must Fix)

1. **SOCKS5 proxy support removed**
   - Breaking change: `proxy: "socks5://..."` configs will fail
   - README updated but no migration guide for existing users
   - **Action**: Document migration or add feature flag for SOCKS5

2. **Test coverage gap**
   - 6781/6790 tests passed - which 9 failed?
   - Need to verify failures are pre-existing, not introduced by this PR
   - **Action**: Provide list of failing tests with justification

3. **Timeout handling semantics changed**
   - curl `--max-time` = total operation timeout
   - std.http timeout = per-request stage timeout
   - **Action**: Verify behavior matches or document difference

4. **Lost mid-flight cancellation**
   - `CancelWatcherCtx` removed (SIGTERM killing)
   - `checkInterrupted()` only checks flag before request starts
   - Long-running requests cannot be cancelled mid-flight
   - **Action**: Document limitation or add cancellation support

### Code Quality

✅ Clean error renaming (Curl* → Http*)
✅ Test injection hooks for mocking
⚠️ No rollback plan if std.http has regressions
⚠️ Missing feature parity with curl (SOCKS5, some TLS options)

### Security

✅ Removes subprocess attack surface
✅ Uses native TLS
⚠️ std.http TLS behavior may differ from curl's
⚠️ Proxy validation needs review

### Recommendations

1. Add feature flag to opt-in to native HTTP
2. Consider gradual rollout (one module at a time)
3. Provide performance benchmarks vs curl
4. Add rollback strategy

This refactor is well-intentioned but needs the above resolved before merge.

Issue #882, opened 2026-05-03
sandbox: default to Landlock on Linux, stop probing external tools at startup


mark-os:
## Problem

The current `sandbox.backend: "auto"` default probes for external sandbox tools at startup by spawning child processes:

1. `firejail --quiet --noprofile /bin/sh -c "exit 0"`
2. `bwrap --ro-bind /usr /usr ... /bin/sh -c "exit 0"`
3. `docker info`

This has caused recurring issues:

- **#791** — Shell broken on physical Linux because auto-detect picked a non-functional sandbox
- **#799** — Docker sandbox broken due to uninitialized mount args; any macOS user with Docker hits this
- **#167 / #289** — Users confused why curl/wget/network commands fail (Docker sandbox uses `--network none`)
- **Zombie processes** — When firejail/bwrap/docker are not installed, `fork()+execve()` fails and leaves zombie children due to a Zig 0.16 stdlib issue where `processSpawnPosix` does not call `waitpid()` on exec failure

The auto-detect priority (firejail → bubblewrap → docker → noop) also means Docker can silently win on systems where it is installed but firejail/bwrap are not. This wraps every shell tool invocation in `docker run --rm --network none alpine:latest`, which:

- Pulls `alpine:latest` on first use (slow, requires network)
- Adds ~200-500ms container startup overhead per command
- Breaks network access (`--network none`)
- Hides the host filesystem outside the workspace mount
- Requires a running Docker daemon

None of this is disclosed to the user during `onboard` or shown in `doctor` output beyond "sandbox: docker".

## Proposal

**Default to Landlock on Linux.** Landlock is a kernel-native LSM (available since Linux 5.13, ~2021) that lets a process restrict its own filesystem access via syscalls — no external tools, no forks, no container overhead, no dependencies.

This aligns with nullclaw's core principles:

| Principle | Landlock | Current auto-detect |
|---|---|---|
| Zero dependencies | ✅ In-kernel | ❌ Requires firejail/bwrap/docker |
| No startup overhead | ✅ Single syscall to detect | ❌ 3-4 fork+exec probes |
| No surprising side effects | ✅ Only restricts filesystem paths | ❌ Docker breaks networking unexpectedly |
| Works on $5 hardware | ✅ Any Linux ≥5.13 | ❌ Docker needs resources; firejail/bwrap may not be packaged |
| Secure by default | ✅ Kernel-enforced, unprivileged | ✅ (when it works) |

### Suggested changes

[...]

PR #883, opened 2026-05-03
probe: resolve executable before spawning child process

mark-os:
## Summary

Adds a pre-spawn executable resolution check to `runQuietCommand` in `src/security/probe.zig`. Before spawning a child process, we now verify the target executable exists on PATH (or at the given absolute/relative path).

## Problem

A Zig stdlib bug causes failed `execve` calls to leave zombie processes. When `runQuietCommand` was called for sandbox probes (docker, firejail, bwrap) and the executable wasn't installed, the spawn would fail but could leave zombies behind.

## Changes

- `canResolveExecutable()`: walks PATH (or checks absolute/relative paths) to verify the executable exists before spawning
- `fileIsExecutable()`: uses `accessAbsolute` with execute permission check
- Empty argv guard on `runQuietCommand`
- Platform-correct PATH separator (`:` on Unix, `;` on Windows)
- Tests for all new code paths

This is also a strict improvement independent of the Zig bug: sandbox probes no longer attempt to spawn processes for executables that don't exist.

## Testing

- `zig build test --summary all` — all tests pass, 0 leaks
- New tests: empty argv, absolute path resolution, bare name PATH lookup, nonexistent binary rejection

Related: #882

### DIFF
--- a/src/security/probe.zig
+++ b/src/security/probe.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const std_compat = @import("compat");
+const fs_compat = @import("../fs_compat.zig");
 const builtin = @import("builtin");
 
 /// Run a probe command with stdio suppressed and treat exit code 0 as success.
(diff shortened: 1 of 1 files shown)


mark-os:
Rebased onto current main (post-#906 merge).

This PR is complementary to #906. That fix defers sandbox auto-detection to first shell tool use, which avoids zombies at startup — but the underlying issue remains: when `runQuietCommand` spawns a child for a binary that isn't installed, the failed `execve` still leaves zombies. This happens on first shell tool invocation instead of at boot.

This PR fixes the root cause by resolving the executable before spawning, which protects all `runQuietCommand` callers regardless of when they run.

DonPrus (approved):

PR #891, opened 2026-05-05
fix(providers): preserve curl probe transport failures

vernonstinebaker:
## Summary
- preserve curl transport failures in the OpenAI-compatible provider path used by provider health probes
- return `CurlDnsError`, `CurlConnectError`, `CurlTimeout`, `CurlTlsError`, `CurlReadError`, `CurlWriteError`, `CurlWaitError`, and `CurlFailed` directly instead of collapsing them into `CompatibleApiError`
- keep the focused regression test covering the transport-error mapping

## Why
NullClaw provider-health checks still need to distinguish transport failures from API-level failures when probing compatible providers. Collapsing curl transport errors into `CompatibleApiError` hides the real failure mode and causes timeouts and connection failures to be misclassified upstream.

## Refresh Status
- rebased/refreshed onto current `main` on 2026-05-19
- current PR head: `7a97e66`
- no functional changes beyond the original fix; this is a branch refresh only

## Validation
- rebased cleanly onto current `main`
- full `zig build test --summary all` remains red on clean upstream `main` on this machine in unrelated tool/file and other baseline failures, so a green full-suite run is not currently available as a signal for this PR

## Notes
- this remains intentionally separate from the NullHub integration PRs
- I did not find another upstream PR that preserves raw curl transport errors for provider probe classification


vernonstinebaker:
Hey @nullclaw/maintainers — this one's been rebased and is still clean. Small fix that makes provider health probe failures actually diagnosable. Would appreciate a look when you have a moment. Please. :-)

addadi (approved):
Approve with minor suggestions:

1. Add catch-all test case for non-transport errors
2. Consider renaming `mapTransportProbeError` for clarity (e.g., `preserveTransportError`)
3. Extract shared error classification to http_util.zig if more sites need it

Sound fix that addresses real gap in health probe error classification.

vernonstinebaker:
@addadi addressed the follow-ups from your review.

Changes made:
- renamed `mapTransportProbeError` to `preserveProbeTransportError` to make the helper intent explicit
- kept the transport-preservation regression coverage for curl transport failures
- added a separate catch-all regression test to assert non-transport errors still collapse to `error.CompatibleApiError`
- rebased the branch onto current `main`

Validation:
- `zig fmt src/providers/compatible.zig`
- `zig build test --summary all` still fails on this machine after rebase in unrelated upstream areas: `tools/file_*`, `skills.installSkillFromGit`, `sqlite_query`, `web_search`, and one gateway log-expectation test, so there is still no clean full-suite signal available for this PR locally

PR head has been updated with the review follow-up commit.

DonPrus (approved):

PR #892, opened 2026-05-05
test(compat/net): add Windows getAddressList regression tests

### LINKED ISSUE #890: Windows: agent provider HTTP fails with `error: HostResolutionFailed` even with literal-IP base_url (curl path works)
### Description

On Windows, `nullclaw agent` cannot reach any provider chat-completion endpoint -- every call dies with `error: HostResolutionFailed`, surfacing in the gateway log as:

```
error(channel_loop): Agent error: error.AllProvidersFailed (OpenRouter: OpenRouterApiError)
```

The curl-backed HTTP path (`http_util`) works fine on the same machine -- `nullclaw models refresh` happily fetches 372 OpenRouter models, and Telegram channel polling runs without issue. Only the Zig-native `std.http.Client` path used by the agent provider call fails.

### Two HTTP paths diverge on Windows

| Path | Backend | Used by | Result |
|------|---------|---------|--------|
| `http_util` | `curl` subprocess | `models refresh`, `update`, Telegram polling | OK |
| Provider client | Zig `std.http.Client` | `nullclaw agent` chat completions | **HostResolutionFailed** |

`models refresh` log on the same machine, same env:

```
Refreshing model catalog...
  Fetching from openai...
warning(http_util): curl GET failed: exit_code=22 class=other
  [SKIP] openai: curl failed
  Fetching from openrouter...
  [OK] openrouter: 372 models
```

curl-shim trace (1265 entries, all `api.telegram.org`, **zero OpenRouter**) confirms the agent path bypasses curl entirely.

### Ruled out

1. **API key.** Direct `curl https://openrouter.ai/api/v1/chat/completions` with the same key returns 200 and a valid completion.
2. **Model id.** Every model fails identically -- `moonshotai/kimi-k2.6`, `moonshotai/kimi-k2-0905`, `~moonshotai/kimi-latest`, `anthropic/claude-haiku-4.5`, even `auto`. All log `success: false, prompt_tokens: 0` in `llm_token_usage.jsonl`.
3. **System DNS.** `nslookup openrouter.ai` and `Resolve-DnsName openrouter.ai` both return `104.18.2.115` (Cloudflare). Strings dump of `nullclaw.exe` shows correct hardcoded URL `https://openrouter.ai/api/v1` -- not `api.openrouter.ai`.
4. **`base_url` override.** `nullclaw config set models.providers.openrouter.base_url https://openrouter.ai/api/v1` -- still fails. Setting it to literal IPv4 `https://104.18.2.115/api/v1` (bypass DNS entirely) -- **still fails**. Failure is not in URL parsing or DNS-server reachability; it is in the Zig std network stack itself.
5. **`nullclaw doctor`** -> `21 ok, 0 warnings, 0 errors`. **`nullclaw config validate`** -> `OK`.
6. **Cloudflare WARP** running but disconnected. **Proxy env** (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`) all empty.
7. **Shell host.** Reproduced from Git Bash and PowerShell -- identical failure.
8. **Version.** Both `nullclaw 2026.5.4` and `nullclaw-2026.4.17` fail identically -- not a recent regression.
9. Likely related class of bug: #317 (`WebSocket gateway fails on Windows: posix.read() uses NtReadFile instead of ws2_32.recv`). Suspicion is the Zig stdlib network path on Windows uses a code path that does not work for sockets / DNS resolution.

### Steps to reproduce

```bash
cd <instance-data-dir>
NULLCLAW_HOME=<instance-data-dir> nullclaw agent --verbose -m "test"
```

Output:

```
warning(agent): Verbose flag detected, enabling verbose logging
info(memory): memory plan resolved: backend=sqlite retrieval=keyword vector=none rollout=off ...
info(agent): Sending to openrouter...
error: HostResolutionFailed
```

Confirm the curl-backed path works on the same env to prove key/network/env are healthy:

```bash
NULLCLAW_HOME=<instance-data-dir> nullclaw models refresh
# -> "[OK] openrouter: 372 models"
```

Confirm the API key works independently:

```bash
curl -sS -X POST https://openrouter.ai/api/v1/chat/completions \
     -H "Authorization: Bearer $OPENROUTER_API_KEY" \
     -H "Content-Type: application/json" \
     -d '{"model":"moonshotai/kimi-k2.6","messages":[{"role":"user","content":"hi"}],"max_tokens":10}'
# -> 200 with valid completion
```

### Expected behavior

`nullclaw agent` should be able to reach OpenRouter (and any other provider) on Windows when `models refresh` already succeeds against the same endpoint with the same env.

### Suggested fix

Route provider chat completions through the existing `http_util` (curl subprocess) path -- the same one `models refresh` and channel modules already use. That keeps a single working HTTP path on Windows and removes the asymmetry where one endpoint succeeds and another fails for the same host.

Alternatively, if Zig std.http is meant to remain the primary path, the underlying Zig stdlib network call needs the Windows-specific syscall fix (similar in spirit to the WebSocket fix tracked in #317 -- `posix.read` -> `ws2_32.recv`).

### Version

`nullclaw 2026.5.4` (also reproduced on `2026.4.17`)

### OS

Other (Windows 11 Pro, build 10.0.26200; tested from both Git Bash and PowerShell)

fatihaziz:
Refs #890.

#890 -- Windows `getAddressList` was a localhost-only stub, so provider chat completions that pre-resolve a hostname could fail before curl made the request. `main` has since fixed the runtime path with the `getAddressListWindows` resolver (973bfa4, dfc9f3b, d3c7ef7), which uses `std.Io.net` async lookup.

This branch originally proposed an alternative hand-rolled `GetAddrInfoW` resolver. After merging current `main`, that implementation is superseded and has been dropped. What remains is deterministic regression coverage on top of main's resolver, plus a tiny lookup-error mapper helper so the mapping can be tested without live DNS.

## What this PR adds

Coverage in `src/compat/net.zig`:

- `foo.localhost` resolves on Windows through the non-fallback resolver branch. Zig handles the `.localhost` suffix locally, so this avoids external DNS while still catching the old exact-`localhost` stub.
- `localhost` keeps the Windows fast-path and returns a single loopback `AF_INET` address.
- Invalid Windows hostnames map to `error.UnknownHostName` rather than `error.Unexpected` or silent success.
- Windows lookup errors (`UnknownHostName`, `NoAddressReturned`, DNS parse failures, system resources, and fallback unexpected errors) map deterministically to the public `GetAddressListError` surface.

Test determinism cleanup in `src/security/docker.zig`:

- Docker sandbox availability returns `false` in unit-test builds instead of spawning `docker info`.
- Runtime behavior is unchanged outside tests: Docker availability still probes daemon reachability with `docker info`.
- This keeps the pre-push `zig build test --summary all` hook from depending on local Docker Desktop or CLI plugins.

Current `main` has been merged into the branch.


abaynp (approved):

fatihaziz:
This PR is ready for review. After merging current `main` (which independently fixed #890 via `getAddressListWindows`), the diff is now scoped to three Windows-gated regression tests for `getAddressList` -- numeric resolution, the `localhost` fast-path, and unresolvable-host error mapping. CI is green on Linux/macOS/Windows. Happy to fold these into the existing windows test if fewer test blocks are preferred. Could a maintainer take a look when convenient?

DonPrus (approved):

Issue #941, opened 2026-05-31
Agent-type cron jobs don't spawn a subprocess — Telegram delivery never happens


weissfl:
### Description

When a scheduled task is created via `schedule` with `job_type: "agent"` and correct delivery parameters (`delivery_mode: "always"`, `delivery_channel: "telegram"`), the job is marked as completed but the agent subprocess is never started. No message arrives in Telegram.

### Steps to reproduce

1. Create a task:
```
schedule once 30s prompt:"say exactly one word: test"
```
2. Wait for it to execute (30 seconds).
3. Check observable logs (`/tmp/nullclaw-observability.jsonl`).

### Expected behavior

After `cron_job_start` there should be an `agent_start` event — the agent subprocess processes the prompt and delivers the result via `deliverResult`.

### Actual behavior

- `cron_job_start` is written to observable.
- The task disappears from `cron.json` (completed).
- **No `agent_start` follows** — the subprocess is never launched.
- **No message arrives in Telegram.**
- No errors in `daemon.log`.

### Analysis

Across the entire observable log (3074 lines) no successful cron-to-agent delivery chain exists: no `agent_start` is ever preceded by a `cron_job_start`. All `agent_start` entries with `channel:"telegram"` belong to direct user messages.

The failure point is in `scheduler.zig` — the call to `runAgentJob()` either never happens or fails before writing an observable event.

Additionally, observable contains binary garbage in the `channel` field of `agent_start` (e.g. `[80,1,137,161,255,255,0,0]`) — likely a use-after-free in serialization for `deliverViaMainAgent`.

### Environment

- **Hardware:** Orange Pi Zero 3, aarch64
- **OS:** Armbian 26.8.0 / Ubuntu 26.04 (Resolute Raccoon), kernel 6.18.31-current-sunxi64
- **NullClaw:** version `2026.5.29`, ELF 64-bit aarch64, statically linked, running as `nullclaw gateway`
- **Channel:** Telegram (bot_account: main)
- **Installation:** `~/.local/bin/nullclaw` (prebuilt binary, not compiled from source)
- **Memory footprint:** ~47 MB in `~/.nullclaw/`
- **Service manager:** `systemctl --user` (process is running, but `systemctl --user cat nullclaw` returned no unit file)

### Test task parameters

```
id: agent-once-5
job_type: agent
session_target: isolated
delete_after_run: false
delivery_mode: always
delivery_channel: telegram
delivery_account_id: main
delivery_to: [redacted]
[...]

PR #954, opened 2026-06-13
Fix: one-shot cron jobs silently fail to deliver messages (use-after-free in OutboundMessage.channel)

vernonstinebaker:
## Problem

GitHub issue #941: agent-type "schedule once" cron jobs silently fail to deliver messages via any channel (Telegram, Mattermost, etc). The job executes, gets removed from `cron.json`, but no output appears in the channel.

Root cause: **use-after-free** on the `OutboundMessage.channel` pointer in `src/bus.zig`.

## Timeline

1. `tick()` in `cron.zig` fires a scheduled job and captures the result into a `Payload{.agent_job_result}`
2. The payload is pushed onto an async bus queue — `msg.channel` is a borrowed pointer into `delivery.channel`
3. `freeJobOwned()` immediately frees `delivery.channel` (and the entire delivery struct)
4. The outbound dispatcher in `dispatch.zig` reads `msg.channel` — it's now a dangling pointer into freed memory
5. `parseTarget()` reads garbage or null → delivery silently fails

Recurring jobs are unaffected because their delivery strings are recreated each cycle and persist until the next tick.

## Fix

Change `OutboundMessage.channel` from a borrowed pointer to an owned string — `dupe` it in all factory functions, free it in `deinit`. This matches the existing ownership model for `chat_id`, `content`, and `account_id`.

### Files changed

- `src/bus.zig` — 5 factory functions (`makeOutboundWithStage`, `makeOutboundWithAccountStage`, `makeOutboundWithChoices`, `makeOutboundWithAccountChoices`, `makeOutboundWithMedia`): dupe `channel` in each; update `deinit` to free it; update comment explaining ownership

### Verification

- Built `zig build -Doptimize=ReleaseSmall` — clean, 0 warnings
- `zig build test --summary all` — 7351 pass, 0 new failures, 2 pre-existing leaks (unrelated)
- **Live production test**: scheduled a one-shot job via the schedule tool → message delivered to Mattermost Workstream channel → confirmed working

## Notes

This fix is channel-agnostic — it resolves delivery for Telegram, Mattermost, and any other channel. The bug is specific to one-shot ("schedule once") jobs; recurring jobs were never affected.


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork main (`4386aa76`) — deployed in production from that fork. Keeping the PR open in case upstream revives.

PR #963, opened 2026-06-18
fix(channels): document and harden Weixin iLink QR auth

### LINKED ISSUE #817: Does nullclaw support WeChat QR code login?
I would like to confirm whether nullclaw supports WeChat QR code login using a QR code scan.\n\nIf it is already supported, please share the configuration or documentation. If not, I would appreciate knowing whether this is planned.\n\nThanks.

vernonstinebaker:
## Summary

Documents the Weixin iLink Bot QR authorization flow in the English and Chinese channel guides, and hardens the existing implementation uncovered during review.

Ref: #817 | Closes #817

## What changed

- Clarifies that `weixin` is the iLink Bot transport and is distinct from the signed-webhook `wechat` Official Account channel and WeCom.
- Makes `nullclaw auth login/status/logout weixin` preserve inline, multi-account, and legacy-array configs without exposing token material.
- Persists proxy and redirected base URL settings, validates credential-bearing HTTPS/proxy URLs, and exits non-zero when credentials cannot be saved.
- Makes empty inbound allowlists fail closed across the affected channel paths; Tencent sender IDs use exact case-sensitive matching.
- Fixes partial-allocation/OOM cleanup in Weixin parsing and login responses.
- Fixes leaked routed gateway session keys across webhook channel handlers.
- Converts stale WeChat/Max/Teams config examples to canonical `accounts` paths and corrects stale IRC `channel` examples.
- Keeps the intentional, documented MaixCam empty-allowlist exception unchanged.

## Validation

- `zig build test --summary all` — 7,402 passed, 9 skipped, zero leaks
- GitHub CI: Linux x86_64, macOS aarch64, and Windows x86_64 all passed

## Compatibility note

For the affected allowlist-based channels, an empty `allow_from` now denies inbound messages. Use explicit sender IDs, or `"*"` only for an intentionally open bot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,7 +431,7 @@ Config: `~/.nullclaw/config.json` (created by `onboard`)
           "host": "irc.libera.chat",
           "port": 6697,
           "nick": "nullclaw",
-          "channel": "#nullclaw",
+          "channels": ["#nullclaw"],
           "tls": true,
           "allow_from": ["user1"]
         },
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -527,6 +527,53 @@ WeChat notes:
 - `allow_from` should list trusted OpenIDs. Keep it explicit; do not rely on an empty allowlist for privacy.
 - Build with `-Dchannels=wechat` (or `-Dchannels=all`) if your binary was compiled without the WeChat channel.
 
+#### Weixin — Personal WeChat Account (QR Code Login)
+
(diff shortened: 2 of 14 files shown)


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork main (`ced302b7`) — deployed in production from that fork. Keeping the PR open in case upstream revives.

PR #966, opened 2026-06-19
fix(http): secure buffered curl fallback on Android

vernonstinebaker:
## Problem

On `aarch64-linux-android` (Termux), Zig 0.16's stdlib HTTP path can fail DNS resolution with `error.NameServerFailure`, while the installed curl resolves the same endpoints successfully. The original patch routed only part of the traffic through curl and did not preserve the full `std.http` request/response contract.

## What changed

- Route every buffered `ProxyHttpClient.fetch` call site through an Android curl adapter; non-Android `ProxyHttpClient.fetch` behavior remains on `std.http`.
- Preserve the exact HTTP method, request body, caller content type, status code, response headers, response-size limit, response writer, timeout, and proxy selection.
- Keep URL, proxy credentials, query tokens, and request headers out of process argv by using exclusive mode-0600 curl config/header files.
- Disable user curl config loading with first-argument `-q`, restrict protocols, suppress curl's implicit form content type, and never return or log raw curl stderr.
- Consolidate the duplicated buffered curl implementations into one executor with bounded streaming and cancellation/process-lifetime handling.
- Preserve explicit/config proxy semantics: environment proxies still honor `NO_PROXY`, while a process-wide config override cannot be silently bypassed by it.
- Follow redirects only for header-free GETs. Requests with caller headers or bodies return redirects unhandled so credentials and methods are not forwarded unsafely.
- Route the old buffered bypasses in auth, channels, vector/memory backends, and SkillForge through `ProxyHttpClient.fetch`.
- Reuse the secure executor in the HTTP request tool and API memory backend. The browser tool now returns redirects for separate validation instead of following an unvalidated redirect into a private address.
- Keep direct streaming curl URLs/proxies out of argv and ignore `.curlrc`; provider timeout arguments and safe direct-connect resolve pins are now honored.

The branch also includes the current `main` via a normal merge commit.

## Security notes

- Sensitive request data is stored only in short-lived mode-0600 files and request bodies are sent through stdin.
- Curl stderr is drained for process safety but not exposed because it can echo credentialed proxy or URL values.
- A configured HTTP/SOCKS proxy remains a separate trust boundary: curl can delegate the proxy's upstream DNS lookup even when a direct connection would use `--resolve` pinning.

## Scope boundaries

- `SseConnection` is a persistent, long-lived transport and cannot use the buffered adapter. Signal's supported default local daemon path does not need DNS; a remote Signal SSE hostname on Android still needs a dedicated persistent curl transport.
- Provider paths that perform the SSRF-safe libc DNS preflight before curl still fail closed on affected Termux systems when that preflight cannot resolve the host. This patch does not bypass that security check; an Android-safe resolver/pinning design is separate work.
- The current revision was not smoke-tested on an Android device in this review environment. The original reporter's earlier revision was tested on Termux, but the hardened revision still needs target-device confirmation.

## Validation

- `zig fmt --check src/`
- `git diff --check`
- `zig build test --summary all` — 13/13 steps, 7363 passed, 9 skipped, 0 failures/leaks
- Pre-push hook reran the full test suite successfully

An Android cross-build was not available locally because the required Android NDK/Zig libc sysroot file was not installed.

## Related upstream

- https://github.com/ziglang/zig/issues/22031
- https://github.com/ziglang/zig/issues/26016

### DIFF
--- a/src/auth.zig
+++ b/src/auth.zig
@@ -380,7 +380,7 @@ pub fn refreshAccessToken(
 
     var aw: std.Io.Writer.Allocating = .init(allocator);
     defer aw.deinit();
-    const result = try client.client.fetch(.{
+    const result = try client.fetch(.{
         .location = .{ .url = token_url },
         .method = .POST,
         .payload = payload,
@@ -506,7 +506,7 @@ pub fn startDeviceCodeFlow(
 
     var aw: std.Io.Writer.Allocating = .init(allocator);
     defer aw.deinit();
-    const result = try client.client.fetch(.{
+    const result = try client.fetch(.{
         .location = .{ .url = device_auth_url },
         .method = .POST,
         .payload = payload,
@@ -599,7 +599,7 @@ pub fn pollDeviceCode(
 
         var aw: std.Io.Writer.Allocating = .init(allocator);
         defer aw.deinit();
-        const result = client.client.fetch(.{
+        const result = client.fetch(.{
             .location = .{ .url = token_url },
             .method = .POST,
             .payload = payload,
--- a/src/channels/dingtalk.zig
+++ b/src/channels/dingtalk.zig
@@ -721,7 +721,7 @@ pub const DingTalkChannel = struct {
     fn postJson(self: *DingTalkChannel, webhook_url: []const u8, body: []const u8) !void {
         var client = try http_util.ProxyHttpClient.init(self.allocator);
         defer client.deinit();
-        const result = client.client.fetch(.{
+        const result = client.fetch(.{
             .location = .{ .url = webhook_url },
             .method = .POST,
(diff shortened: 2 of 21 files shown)


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork main (`679e8f5c`) — deployed in production from that fork. Keeping the PR open in case upstream revives.

vernonstinebaker:
Follow-up after this merges, separate pull request: log the scrubbed provider error body on non-2xx. That is the diagnosability half of #1000. It hooks `curlPostTypedTimed`, which this pull request adds, so it should not be opened against current `main` and should not be folded into this branch. It does not add a tool-capability check.

PR #970, opened 2026-06-29
fix(cli): handle arrow keys in agent REPL

### LINKED ISSUE #865: CLI shows ctrl characters for up/down/left/right keys
### Description

Apparently the CLI breaks native keybinding. Instead, the nullclaw CLI shows CTRL character garbage instead of moving up/down in last command buffer, or moving cursor left right on line. 

### Expected behavior

That up/down, and left/right keys works as expected. 

### Steps to reproduce

Image

### Version

4.19

### OS

Linux

vernonstinebaker:
## Summary
- Add a small allocation-free line editor for the interactive `nullclaw agent` REPL.
- Enable POSIX raw-mode input for TTY sessions so arrow keys, history navigation, cursor movement, backspace/delete, Home/End, and common word-left/right sequences are handled instead of printed as control characters.
- Preserve the existing canonical input fallback for pipes, non-TTYs, WASI, and native Windows.

Fixes #865.

## Validation
- `zig version` -> `0.16.0`
- `zig test src/channels/cli_line_editor.zig` -> 5/5 passed
- `zig build` -> passed
- `git diff --check` -> passed
- `zig build test --summary all` -> failed on existing/unrelated leak in `gateway.test.run returns AddressInUse when port is already bound` after 7257 pass, 9 skip, 9 leaks
- Manual smoke test with `nullclaw-test-arrows agent`: improved behavior; arrow control characters no longer print

## Notes
- This targets the interactive `nullclaw agent` prompt path only.
- WSL should use the POSIX path. Native Windows keeps the existing fallback; native Windows Console key-event handling can be a follow-up if needed.
- Please add a comment/link in issue #865 to this PR so the reporter can smoke test the fix.

---
_Replaces #960 (retargeted from `codex/fix-cli-arrow-keys` to `fix/cli-arrow-keys` per tool-neutral branch naming convention). Same content, branch renamed._


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork main (`3511e32c`) — deployed in production from that fork. Keeping the PR open in case upstream revives.

PR #971, opened 2026-06-29
feat(streaming): native tool calls during SSE streaming

vernonstinebaker:
## Summary

Decouple native tool-call support from the streaming path so that providers which support native tools during streaming can actually emit them.

Previously the agent loop disabled native tools whenever a stream callback was attached, forcing tools into a prompt-injection format that several models could not reliably emit. Worse, the streaming protocol (`StreamChunk` / `StreamChatResult`) had no field to carry tool calls, so even when a provider parsed them from the SSE deltas they were dropped at the agent boundary.

## Background

The streaming implementation predates native tool support in the OpenAI-compatible protocol. When streaming was added, the response protocol was text-delta-only (`StreamChunk{ delta, is_final, token_count }`), and the streaming call site passed `.tools = null`. The explicit `native_tools_enabled = !is_streaming and ...` gate was added later (during a logging refactor) and codified a limitation that had already become structural rather than intentional — streaming tool-call deltas have been a documented part of the OpenAI-compatible protocol for years.

This restores provider-protocol parity for the streaming path.

## Changes

- **providers/root**: add `tool_calls` to `StreamChatResult` and a `supportsStreamingNativeTools()` vtable slot, defaulting to `false` (conservative — providers must opt in).
- **providers/sse**: add a `ToolCallAccumulator` that parses `delta.tool_calls` (OpenAI-compatible format, including parallel calls indexed by `index`) across SSE chunks and emits a complete `tool_calls` slice on the final `StreamChatResult`.
- **providers/{openai,compatible}**: opt in to streaming native tools.
- **providers/{router,reliable}**: delegate the capability conservatively to the underlying provider.
- **providers/anthropic**: explicitly report `false`. Anthropic streaming tool-use uses a different event shape (`input_json_delta` content blocks) and is left as a follow-up to keep this PR focused.
- **agent/root**: gate native tools on `supportsStreamingNativeTools()` when streaming (rather than on `!is_streaming`), and propagate `stream_result.tool_calls` into the agent loop instead of dropping them.

## Tests

- Regression tests asserting native tools are requested during streaming when the provider advertises the capability.
- Regression tests asserting `tool_calls` emitted on the streaming path reach the agent loop (not dropped).
- `ToolCallAccumulator` unit tests for delta accumulation, parallel-indexed calls, and malformed-input safety.

## Validation

```
zig build test --summary all
Build Summary: 13/13 steps succeeded; 7362/7371 tests passed (9 skipped)
```

0 failures, 0 leaks (`std.testing.allocator`).

`zig fmt --check src/` clean.

## Security / Risk

- No change to tool execution policy, allowlists, or sandboxing. The shell tool's existing security gates (command allowlist, redirection policy, risk classification) apply identically to tool calls arriving via the streaming path.
- No secrets, tokens, or payloads are logged.
- Conservative capability default (`false`): only OpenAI-compatible providers opt in. Adding the capability to a provider that does not actually parse `delta.tool_calls` would silently drop tool calls, so the opt-in is intentional per provider.

## Scope

Provider streaming protocol only. No config schema, vtable signatures beyond the new `supportsStreamingNativeTools()` slot, or user-facing setup changes. Anthropic streaming tool-use is explicitly out of scope for this PR.

### DIFF
--- a/src/agent/root.zig
+++ b/src/agent/root.zig
@@ -2022,7 +2022,10 @@ pub const Agent = struct {
             defer prompt_tools_arena.deinit();
             const prompt_tools = try self.filterToolsForPromptText(prompt_tools_arena.allocator());
             const prompt_is_streaming = self.stream_callback != null and self.stream_ctx != null and self.provider.supportsStreaming();
-            const prompt_native_tools_enabled = !prompt_is_streaming and self.provider.supportsNativeTools();
+            const prompt_native_tools_enabled = if (prompt_is_streaming)
+                self.provider.supportsStreamingNativeTools()
+            else
+                self.provider.supportsNativeTools();
 
             const capabilities_section = capabilities_mod.buildPromptSection(
                 self.allocator,
@@ -2200,7 +2203,10 @@ pub const Agent = struct {
 
             const timer_start = std_compat.time.milliTimestamp();
             const is_streaming = self.stream_callback != null and self.stream_ctx != null and self.provider.supportsStreaming();
-            const native_tools_enabled = !is_streaming and self.provider.supportsNativeTools();
+            const native_tools_enabled = if (is_streaming)
+                self.provider.supportsStreamingNativeTools()
+            else
+                self.provider.supportsNativeTools();
             const include_reasoning = self.reasoning_mode != .off;
 
             // Filter tool specs for this turn (arena-owned; may be self.tool_specs directly if no groups).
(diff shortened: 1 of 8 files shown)


vernonstinebaker:
Marking as draft pending **#970** (fix(cli): handle arrow keys in agent REPL).

Both PRs modify overlapping regions of `src/agent/root.zig`. When #970 merges to `main`, this branch should rebase cleanly onto the updated file. Until then, the two conflict when combined on an integration branch.

This PR is complete and validated on its own (`zig build test --summary all` → 7362/7371 passed, 9 skipped, 0 failures, 0 leaks). The draft status reflects only the sequencing dependency, not the readiness of the change.

vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork main (`ce2d54ca`) — deployed in production from that fork. Keeping the PR open in case upstream revives.

PR #979, opened 2026-07-29
feat(memory): add configurable auto-recall, recall_limit, max_context_bytes

### LINKED ISSUE #919: Feature: Allow disabling automatic memory recall (FTS5) per-message
## Problem

`enrichMessageWithRuntime()` runs FTS5 + BM25 recall on **every incoming message**, with hardcoded parameters:

- `DEFAULT_RECALL_LIMIT = 5`
- `MAX_CONTEXT_BYTES = 4000`
- `SCOPED_RECALL_CANDIDATE_LIMIT = 64`
- `GLOBAL_RECALL_CANDIDATE_LIMIT = 64`

There is no way to disable this behavior via config. `auto_save: false` prevents automatic memory writes, but recall still fires on every message — even when the `memories` table is empty, causing unnecessary SQLite queries.

## Why this matters

1. **Unnecessary overhead** — FTS5 search runs even when memory is empty or the user does not need context injection. For high-throughput or professional use, this is wasted cycles.

2. **Noisy context** — Auto-recall can inject irrelevant memories into the prompt, degrading model quality. The user has no way to say "I will manage context myself."

3. **Inconsistent with `auto_save: false`** — If I can disable auto-writes, I should be able to disable auto-reads. Currently it is only half-controllable.

4. **Hardcoded limits** — `recall_limit`, `max_context_bytes`, and candidate limits are compile-time constants. Users with large memory stores or specific context budgets cannot tune them.

## Proposal

Add config options under `memory`:

```json
{
  "memory": {
    "auto_recall": true,
    "recall_limit": 5,
    "max_context_bytes": 4000
  }
}
```

- `auto_recall: false` — skip `enrichMessageWithRuntime()` entirely, no FTS5/LIKE queries
- `recall_limit` — configurable max results (currently hardcoded to 5)
- `max_context_bytes` — configurable context budget (currently hardcoded to 4000)

Manual `memory_recall` tool calls should still work regardless of `auto_recall` setting.

valonmulolli:
## Summary

Adds three new JSON config keys under `memory` to control memory recall behavior:

- `auto_recall` (bool, default `true`) - when `false`, skips memory enrichment entirely (no FTS5/LIKE queries on user messages)
- `recall_limit` (u32, default `5`) - max memory entries injected per request
- `max_context_bytes` (u32, default `4000`) - total context budget in bytes

## Validation

```
zig build test -Dtarget=x86_64-linux-gnu --summary all
-> 7252 pass, 0 fail, 14 skip, 0 leaks
```

Closes #919


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork at `72db0008` — as a semantic re-implementation rather than a cherry-pick, because the memory loader and the agent enrichment call site were reworked after this PR was opened. Key names and semantics are preserved (including `recall_limit: 0` disabling injection). One deliberate divergence, recorded in the fork plan: `auto_recall=false` does not broaden response-cache eligibility — any turn with a memory backend configured stays cache-ineligible per the fork's cache hardening. Docs for the three keys were added in English and Chinese, including the distinction from `memory.search.query.max_results`. Thank you — this closes a real gap for small-model setups (upstream #919).

vernonstinebaker:
The original branch is gone, so this cannot be reopened. The same change is now https://github.com/nullclaw/nullclaw/pull/1001, cherry-picked onto current main. Authorship of the feature commit stays with @valonmulolli.

PR #980, opened 2026-07-29
fix(scheduler): persist paired token to disk during /pair

### LINKED ISSUE #839: bug: bit has no access to scheduler !?

valonmulolli:
## Fixes #839

The `/pair` endpoint generates a token and stores its hash in memory, but never writes it to disk. The cron/schedule tool reads from `{config_dir}/paired_token` to authenticate against gateway admin routes, but since the file never exists, `readPairedToken()` returns `null` and the gateway returns 401.

### `src/gateway.zig` (write path)
- After successful pairing in the `.paired => |token|` handler, persist the token to `{config_dir}/paired_token` encrypted at rest via `SecretStore.encryptSecret()` (ChaCha20-Poly1305)
- Failure to persist is logged as a warning but doesn't break the pairing response — the in-memory token still works for the current session

### `src/cron.zig` (read path)
- Updated `readPairedToken()` to detect encrypted tokens via `SecretStore.isEncrypted()` and decrypt them before returning
- **Backward compatible**: plaintext tokens (from older versions or manual setup) still work unchanged

### Tests
- Roundtrip test: encrypt → write file → read → decrypt → matches original token
- Backward compat test: plaintext token file → read → works without encryption
- gateway integration test: `persistPairedToken()` creates encrypted file that decrypts back

## Verification
- Code compiles cleanly (0 compilation errors from changes)
- Pre-existing linker errors (`R_X86_64_PC64` in `crt1.o`) are a system GCC 16 / Zig 0.16 toolchain issue, unrelated to these changes

### DIFF
--- a/src/cron.zig
+++ b/src/cron.zig
@@ -14,6 +14,7 @@ const signal = @import("channels/signal.zig");
 const Config = @import("config.zig").Config;
 const process_util = @import("tools/process_util.zig");
 const security_policy = @import("security/policy.zig");
+const secrets = @import("security/secrets.zig");
 
 const log = std.log.scoped(.cron);
 const DEFAULT_CRON_SHELL_TIMEOUT_NS: u64 = 60 * std.time.ns_per_s;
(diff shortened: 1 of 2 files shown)


vernonstinebaker:
This PR and #959 both fix #839 with the same approach (persist the paired token to `<config_dir>/paired_token`, encrypted at rest via the shared `SecretStore`, written on `/pair`). For anyone picking this issue up: #959's implementation is merged and running in the `vernonstinebaker/nullclaw` fork main (`bce0ba8c`) — it additionally decrypts `enc2:` blobs in the scheduler tool's `readPairedToken` for legacy-token compatibility and clears the file on `/logout`. The two implementations are close enough that cross-checking may save a rework if this lands upstream.

PR #983, opened 2026-08-03
fix(providers): use pinned curl path for proxied requests

ArcanePivot:
## Summary
- route non-streaming provider POSTs through the existing secure curl path when a safe pinned resolve entry is available
- keep credential headers out of argv by reusing the mode-0600 temporary header file
- preserve the std.http fallback when no pinned resolve entry is available
- pass the configured provider timeout to curl instead of discarding it

## Why
On MIPS32 OpenWrt behind an explicit HTTP proxy, streaming provider requests worked because they use curl, while both DeepSeek non-streaming models consistently failed through the std.http proxy path with `CompatibleApiError`. This also prevented the non-streaming reliability/model-fallback chain from being usable on that target.

## Validation
- `zig fmt src/providers/helpers.zig`
- MIPS32r2 static build: `zig build -Dtarget=mipsel-linux-musleabi -Dcpu=mips32r2 -Doptimize=ReleaseSmall -Dembedded_wasm3=false`
- real-device forced non-streaming calls succeeded for both `deepseek-v4-flash` and `deepseek-chat` through the HTTP proxy
- full test build compiled and ran 7,381 tests: 7,369 passed, 10 skipped; two pre-existing environment-sensitive DNS/curl tests failed on the macOS host

### DIFF
--- a/src/providers/helpers.zig
+++ b/src/providers/helpers.zig
@@ -502,12 +502,33 @@ pub fn convertToolsResponses(buf: *std.ArrayListUnmanaged(u8), allocator: std.me
 /// HTTP POST with optional LLM timeout (seconds). 0 = no limit.
 /// Automatically reads proxy from HTTPS_PROXY, HTTP_PROXY, or ALL_PROXY environment variables.
 pub fn curlPostTimed(allocator: std.mem.Allocator, url: []const u8, body: []const u8, headers: []const []const u8, timeout_secs: u64) ![]u8 {
-    _ = timeout_secs;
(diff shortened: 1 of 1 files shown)


vernonstinebaker:
Superseded by #966.

#966 routes these provider POSTs through the secure curl path. It pins DNS, honors the configured timeout, applies `HTTPS_PROXY` / `HTTP_PROXY` / `ALL_PROXY` when the caller does not pass a proxy, and keeps credentials out of argv. A non-2xx response still comes back as `error.HttpStatusError`. This patch calls `curlPostWithProxyAndResolve`, and curl exits 0 on HTTP 401 and 500, so those responses would be treated as success.

Nothing in this PR needs to land separately. Leave it open until #966 merges, then close it.

PR #984, opened 2026-08-05
fix(channels): let poll failures age out a dead polling thread

### LINKED ISSUE #972: telegram channel stop respond after some idle time
### Description

`Telegram` channel die away at next morning after idle for a night or more time.

But It seems `nullclaw` work well at backend:
```
[ec2-user@ip-<...> ~]$ nullclaw agent -m "ping"
info(memory): memory plan resolved: backend=hybrid retrieval=keyword vector=none rollout=off hygiene=true snapshot=false cache=false semantic_cache=false summarizer=false sources=1
pong
[ec2-user@ip-<...> ~]$ nullclaw service status
Service state: active
Unit: /home/ec2-user/.config/systemd/user/nullclaw.service
```

### Expected behavior

`Telegram` channel should respond as usual.

### Steps to reproduce

- deploy `nullclaw` in AWS Lightsail instance (AWS Linux 2023)
- start `nullclaw` service
- configure `telegram` channel
- talk with `nullclaw` through `telegram`
- idle for a night or more time
- `telegram` die away whill `nullclaw` work well at backend

### Version

v2026.5.29

### OS

Linux

raskevichai:
Closes #972.

## Summary

Telegram and Matrix channels go silent after an idle night while `nullclaw agent` still answers, and only a full gateway restart brings them back. The supervisor is supposed to catch exactly this, but it was structurally blind to it.

## Root cause

`supervisionLoop` decides a polling thread is dead by age:

```zig
const stale = (now - last) > STALE_THRESHOLD_SECS;  // 600s
```

But every polling loop refreshed that timestamp from inside its own failure path:

```zig
const messages = tg_ptr.pollUpdates(allocator) catch |err| {
    log.warn("Telegram poll error: {}", .{err});
    loop_state.last_activity.store(std_compat.time.timestamp(), .release);  // <-
    std_compat.thread.sleep(5 * std.time.ns_per_s);
    continue;
};
```

So a poll that fails forever kept `now - last` near zero and `stale` never became true. That left `healthCheck()` as the only signal, and it is a different request: a short `getMe` / `whoami` that keeps returning 200 while the long-poll connection is dead. The channel was functionally dead and reported healthy.

All five polling loops had the write: `runTelegramLoop`, `runSignalLoop`, `runWeixinLoop`, `runMatrixLoop`, `runMaxLoop`.

The reporter in #972 and a second operator in the thread both landed on this line independently.

## Changes

**1. Failure path no longer touches `last_activity`** (all five loops). The timestamp now tracks successful polls only, which is the invariant the 600s threshold was written against. A thread that cannot poll ages out and the supervisor restarts it.

**2. Exponential backoff on poll failure**, shared via `nextPollBackoffNs` (1s, doubling, 30s cap, reset to 1s on success). Telegram, Signal, Weixin and Matrix previously retried every 5s indefinitely regardless of how long the outage lasted. `runMaxLoop` already had this backoff written inline and now uses the shared helper instead.

**3. `ChannelManager.isPollingStale(now, last)`** extracted from `supervisionLoop`, with `STALE_THRESHOLD_SECS` hoisted to module scope so the rule is unit-testable. No behavior change.

## Tests

Four regression tests, each commented with the issue number:

- `nextPollBackoffNs doubles up to the cap` - 1s to 2s to 4s, saturates at 30s.
- `nextPollBackoffNs saturates instead of overflowing` - guards the `*|` saturating multiply.
- `isPollingStale flags a polling thread past the threshold` - boundary at exactly 600s and 601s.
- `isPollingStale sees a loop state left untouched by failing polls` - a `TelegramLoopState` whose timestamp is not refreshed does cross the threshold.

Sanity check that the tests actually bite: I broke both helpers (removed the backoff cap, multiplied the stale threshold by 100), reran, and got **5 failures** instead of 0 - all four new tests. Restoring produced a byte-identical tree.

## Verification

```
zig fmt --check src/          # clean for both changed files
zig build test --summary all  # 7379/7393 passed, 14 skipped, 0 failed
```

## Scope

Deliberately not included:

- **Recovery from `gave_up`.** `SupervisedChannel` still has no transition out of `gave_up` after `max_restarts`, so a long outage can still permanently retire a channel. That is a real second bug, but it changes the supervisor state machine and belongs in its own PR.
- **Wiring `channel_initial_backoff_secs` / `channel_max_backoff_secs`.** These exist under `reliability` in config but are not read by the polling loops. This PR keeps the constants hardcoded (matching the values `runMaxLoop` already used) rather than mixing a config-schema change into a bug fix.
- **Making `poll_timeout` configurable**, requested in the thread for the same reason.

## Note on test coverage

The unit tests cover the two extracted rules, not the loop bodies themselves - `runTelegramLoop` and friends require a live HTTP endpoint to exercise. The invariant that the failure path must not write `last_activity` is documented in the doc comment on `nextPollBackoffNs` so the reason survives future edits.

### DIFF
--- a/src/channel_loop.zig
+++ b/src/channel_loop.zig
@@ -1245,6 +1245,19 @@ fn matrixRoomPeerId(reply_target: ?[]const u8) []const u8 {
     return reply_target orelse "unknown";
 }
 
+/// Backoff bounds shared by every polling loop's failure path.
+pub const POLL_BACKOFF_INITIAL_NS: u64 = std.time.ns_per_s;
+pub const POLL_BACKOFF_MAX_NS: u64 = 30 * std.time.ns_per_s;
+
+/// Next delay after a failed poll: doubles up to POLL_BACKOFF_MAX_NS.
+///
+/// Poll failures deliberately leave `last_activity` untouched. It is the only
+/// signal that survives a long-poll which fails forever — `healthCheck()` issues
+/// a separate short request that keeps succeeding while the poll is dead.
+pub fn nextPollBackoffNs(current_ns: u64) u64 {
+    return @min(current_ns *| 2, POLL_BACKOFF_MAX_NS);
+}
+
 // ════════════════════════════════════════════════════════════════════════════
 // TelegramLoopState — shared state between supervisor and polling thread
 // ════════════════════════════════════════════════════════════════════════════
@@ -1568,15 +1581,18 @@ pub fn runTelegramLoop(
         1;
     const enable_parallel = max_parallel_tasks > 1;
 
+    var poll_backoff_ns: u64 = POLL_BACKOFF_INITIAL_NS;
+
     while (!loop_state.stop_requested.load(.acquire) and !daemon.isShutdownRequested()) {
         const messages = tg_ptr.pollUpdates(allocator) catch |err| {
             log.warn("Telegram poll error: {}", .{err});
-            loop_state.last_activity.store(std_compat.time.timestamp(), .release);
-            std_compat.thread.sleep(5 * std.time.ns_per_s);
+            std_compat.thread.sleep(poll_backoff_ns);
+            poll_backoff_ns = nextPollBackoffNs(poll_backoff_ns);
             continue;
         };
 
         // Update activity after each poll (even if no messages)
+        poll_backoff_ns = POLL_BACKOFF_INITIAL_NS;
         loop_state.last_activity.store(std_compat.time.timestamp(), .release);
 
         for (messages) |msg| {
@@ -1912,16 +1928,18 @@ pub fn runSignalLoop(
     loop_state.last_activity.store(std_compat.time.timestamp(), .release);
 
     var evict_counter: u32 = 0;
+    var poll_backoff_ns: u64 = POLL_BACKOFF_INITIAL_NS;
 
     while (!loop_state.stop_requested.load(.acquire) and !daemon.isShutdownRequested()) {
         const messages = sg_ptr.pollMessages(allocator) catch |err| {
             log.warn("Signal poll error: {}", .{err});
-            loop_state.last_activity.store(std_compat.time.timestamp(), .release);
-            std_compat.thread.sleep(5 * std.time.ns_per_s);
+            std_compat.thread.sleep(poll_backoff_ns);
+            poll_backoff_ns = nextPollBackoffNs(poll_backoff_ns);
             continue;
         };
 
         // Update activity after each poll (even if no messages)
+        poll_backoff_ns = POLL_BACKOFF_INITIAL_NS;
         loop_state.last_activity.store(std_compat.time.timestamp(), .release);
 
         for (messages) |msg| {
@@ -2040,15 +2058,17 @@ pub fn runWeixinLoop(
     loop_state.last_activity.store(std_compat.time.timestamp(), .release);
 
     var evict_counter: u32 = 0;
+    var poll_backoff_ns: u64 = POLL_BACKOFF_INITIAL_NS;
 
     while (!loop_state.stop_requested.load(.acquire) and !daemon.isShutdownRequested()) {
         const messages = wx_ptr.pollMessages(allocator) catch |err| {
             log.warn("Weixin poll error: {}", .{err});
-            loop_state.last_activity.store(std_compat.time.timestamp(), .release);
-            std_compat.thread.sleep(5 * std.time.ns_per_s);
+            std_compat.thread.sleep(poll_backoff_ns);
+            poll_backoff_ns = nextPollBackoffNs(poll_backoff_ns);
             continue;
         };
 
+        poll_backoff_ns = POLL_BACKOFF_INITIAL_NS;
         loop_state.last_activity.store(std_compat.time.timestamp(), .release);
 
         for (messages) |msg| {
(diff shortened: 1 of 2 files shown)


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork at `4d0b5cfd` — applied verbatim (clean apply), including your regression tests on the extracted `isPollingStale` / `nextPollBackoffNs` seams. Added two edge tests on top: `isPollingStale` with a backwards clock step, and `nextPollBackoffNs(0)` pinned so the failure-path contract cannot change silently. Verified while landing: gateway-loop channels (Mattermost, Discord) use the health-check supervision path and are untouched. Thank you — silent-death-after-idle-night was a nasty one for Telegram/Matrix operators.

src/channel_loop.zig
```diff
@@ -1245,6 +1245,19 @@ fn matrixRoomPeerId(reply_target: ?[]const u8) []const u8 {
     return reply_target orelse "unknown";
 }
 
+/// Backoff bounds shared by every polling loop's failure path.
+pub const POLL_BACKOFF_INITIAL_NS: u64 = std.time.ns_per_s;
+pub const POLL_BACKOFF_MAX_NS: u64 = 30 * std.time.ns_per_s;
```
vernonstinebaker: These bounds are only used in this file. Keeping them private matches the Copilot note and leaves `nextPollBackoffNs` as the callable seam.

```suggestion
const POLL_BACKOFF_INITIAL_NS: u64 = std.time.ns_per_s;
const POLL_BACKOFF_MAX_NS: u64 = 30 * std.time.ns_per_s;
```

src/channel_loop.zig
```diff
@@ -2582,6 +2602,20 @@ test "TelegramLoopState stop_requested toggle" {
...
+    // is down for hours stops hammering the API.
+    try std.testing.expectEqual(@as(u64, 2 * std.time.ns_per_s), nextPollBackoffNs(POLL_BACKOFF_INITIAL_NS));
+    try std.testing.expectEqual(@as(u64, 4 * std.time.ns_per_s), nextPollBackoffNs(2 * std.time.ns_per_s));
+    try std.testing.expectEqual(POLL_BACKOFF_MAX_NS, nextPollBackoffNs(16 * std.time.ns_per_s));
+    try std.testing.expectEqual(POLL_BACKOFF_MAX_NS, nextPollBackoffNs(POLL_BACKOFF_MAX_NS));
+}
+
+test "nextPollBackoffNs saturates instead of overflowing" {
+    try std.testing.expectEqual(POLL_BACKOFF_MAX_NS, nextPollBackoffNs(std.math.maxInt(u64)));
+}
```
vernonstinebaker: Pin the zero input. Callers start at `POLL_BACKOFF_INITIAL_NS`, so `0` is not a live state; doubling should keep it `0` rather than jumping to the initial delay.

```suggestion
test "nextPollBackoffNs saturates instead of overflowing" {
    try std.testing.expectEqual(POLL_BACKOFF_MAX_NS, nextPollBackoffNs(std.math.maxInt(u64)));
}

test "nextPollBackoffNs documents the zero input" {
    // Callers initialize to POLL_BACKOFF_INITIAL_NS; a zero input is not a real
    // state, and doubling preserves it. Pinned so a future "fix" does not
    // silently change the failure-path contract.
    try std.testing.expectEqual(@as(u64, 0), nextPollBackoffNs(0));
}
```

src/channel_manager.zig
```diff
@@ -32,6 +32,10 @@ const Channel = channels_mod.Channel;
 
 const log = std.log.scoped(.channel_manager);
 
+/// How long a polling thread may go without a successful poll before the
+/// supervisor restarts it.
+pub const STALE_THRESHOLD_SECS: i64 = 600;
```
vernonstinebaker: Same visibility note as the backoff bounds: the threshold is only used in this file, including its tests.

```suggestion
const STALE_THRESHOLD_SECS: i64 = 600;
```

src/channel_manager.zig
```diff
@@ -491,6 +502,30 @@ pub const ChannelManager = struct {
...
+    // failure path no longer writes last_activity, so the timestamp stays at the
+    // last successful poll and crosses the threshold.
+    var loop_state = channel_loop.TelegramLoopState.init();
+    const last_success: i64 = 1_700_000_000;
+    loop_state.last_activity.store(last_success, .release);
+
+    const observed = loop_state.last_activity.load(.acquire);
+    try std.testing.expectEqual(last_success, observed);
+    try std.testing.expect(ChannelManager.isPollingStale(last_success + 601, observed));
+}
```
vernonstinebaker: A `last_activity` newer than `now` (clock step back, or a poll finishing between the two reads) must not count as stale.

```suggestion
    try std.testing.expect(ChannelManager.isPollingStale(last_success + 601, observed));
}

test "isPollingStale tolerates a backwards clock step" {
    // A last_activity newer than `now` (NTP step-back, or a poll completing
    // between the two reads) must not read as stale.
    const last: i64 = 1_700_000_000;
    try std.testing.expect(!ChannelManager.isPollingStale(last - 5, last));
    try std.testing.expect(!ChannelManager.isPollingStale(last, last));
}
```

vernonstinebaker (commented):
@raskevichai Four apply-on-GitHub suggestions below. Committing them keeps this PR yours, so I can approve and merge afterward without pushing to the branch.

Two of them are the edge tests already running on the fork intake of this patch: `nextPollBackoffNs(0)` stays `0`, and a backwards clock step is not stale. The other two drop `pub` from the backoff and staleness constants, which is the Copilot note. Nothing here changes the failure-path behavior.

Please use **Add suggestion to batch** on each, then **Commit suggestions**. I will stay off the branch.

PR #985, opened 2026-08-05
fix(runtime): give the agent turn path a 16 MiB stack

### LINKED ISSUE #976: SIGSEGV on every inbound Telegram message — inbound worker thread spawned with a ~512 KB stack overflows
## Description

On aarch64 Linux, nullclaw **v2026.5.29** segfaults (SIGSEGV) on **every** inbound Telegram message. As a `nullclaw gateway` systemd service (`Restart=always`) it crash-loops — each message kills the process, it restarts, and the message is dropped, so the user never gets a reply (`NRestarts` climbs by one per message). `nullclaw channel start` crashes outright on the first message.

**Root cause:** the inbound message → agent pipeline runs on a worker thread spawned with a far-too-small stack. By interposing `pthread_attr_setstacksize` via `LD_PRELOAD`, I observed the inbound worker threads requesting stacks of **524288 bytes (512 KB)** and 2097152 bytes (2 MB), while nullclaw's other threads get 16 MB. The pipeline's on-stack usage exceeds 512 KB, so `$sp` runs into the `PROT_NONE` guard page → SIGSEGV.

**Evidence it is a stack overflow (not recursion or heap corruption):**

- At the crash, `$sp` sits inside a `PROT_NONE` guard mapping immediately *below* the worker's stack region (`info proc mappings`): guard `…b30000–…b40000`, stack `…b40000–…cc0000` (`0x180000` total incl. TLS/metadata). SP had crossed below the stack into the guard.
- Deterministic, identical crash in both `channel start` and `gateway`: same `pc=0x19dba64` (inlined bulk `str q0,[x0]` copy, `x2=528`), same `x30=0x13bef44`.
- The return address appears **0 times** anywhere on the stack → not infinite recursion.
- `nullclaw agent -m "hi"` works perfectly — it runs on the 8 MB main-thread stack. So config / provider / LLM path are healthy; only the small-stack worker path crashes.

**Workaround that fully fixes it:** an `LD_PRELOAD` shim intercepting `pthread_attr_setstacksize` / `pthread_create` to clamp every stack to ≥16 MB. With it loaded, messages are processed normally and the bot replies (verified live; `NRestarts` stays 0). This strongly implies the fix is simply to spawn the inbound worker with a larger `stack_size` (match the 16 MB used elsewhere, or size it to the pipeline's real requirement).

**Environment:** aarch64 (Raspberry Pi, Linux `6.12.47+rpt-rpi-v8`, Debian), glibc ≥ 2.34, dynamically linked. The binary imports `pthread_create` / `pthread_attr_setstacksize@GLIBC_2.34` (Zig libc pthread backend), so `pthread_attr_setstacksize` is the exact call site that sets the worker stack.

## Expected behavior

Inbound messages are processed without crashing. The inbound/dispatcher worker thread should get enough stack — ideally the same 16 MB the other worker threads get, or a value sized to the message→agent pipeline's real peak usage.

## Steps to reproduce

1. Configure a Telegram channel (bot token + `allow_from`).
2. Run `nullclaw gateway` (or `nullclaw channel start`).
3. Send any message to the bot.
4. → SIGSEGV. Under systemd (`Restart=always`) the process crash-loops (`NRestarts` +1 per message); the message is never answered.

Confirm it is the worker stack with a minimal LD_PRELOAD shim:

```c
// stackfix.c — clamp every pthread stack to >= 16 MB
#define _GNU_SOURCE
#include <pthread.h>
#include <dlfcn.h>
#include <stdio.h>
int pthread_attr_setstacksize(pthread_attr_t *a, size_t sz) {
    static int (*real)(pthread_attr_t*, size_t);
    if (!real) real = dlsym(RTLD_NEXT, "pthread_attr_setstacksize");
    size_t min = 16UL<<20;
    if (sz < min) { fprintf(stderr, "[stackfix] bump %zu -> %zu\n", sz, min); sz = min; }
    return real(a, sz);
}
```

```sh
cc -shared -fPIC -O2 -o stackfix.so stackfix.c -ldl
LD_PRELOAD=$PWD/stackfix.so nullclaw channel start
# -> [stackfix] bump 524288 -> 16777216 ; no crash, bot replies normally.
```

## Version

v2026.5.29 (Homebrew bottle)

## OS

Linux

raskevichai:
Closes #976.

## Root cause

`SESSION_TURN_STACK_SIZE` was aliased to `HEAVY_RUNTIME_STACK_SIZE`:

```zig
pub const SESSION_TURN_STACK_SIZE: usize = HEAVY_RUNTIME_STACK_SIZE;  // 2 MiB
```

That constant sizes every thread which runs `SessionManager.processMessage*()` / `Agent.turn()`. The turn path is the deepest stack in the runtime (channel decode -> session -> agent loop -> provider -> tool dispatch), and aarch64 frames are larger than x86_64 ones, so 2 MiB was not enough there: the stack pointer crossed into the `PROT_NONE` guard page below the thread stack and the process died on **every** inbound message.

`nullclaw agent -m "hi"` was never affected because it runs on the 8 MiB main-thread stack, which is why the config, provider and LLM paths all looked healthy.

## Why 16 MiB

The reporter established the mechanism from both directions:

- At the crash `$sp` was inside the guard mapping immediately below the worker stack, and the return address appeared **zero** times on the stack - so a stack overflow, not runaway recursion.
- An `LD_PRELOAD` shim clamping every `pthread_attr_setstacksize` call to 16 MiB made the bot reply normally with `NRestarts` staying at 0.

Their `LD_PRELOAD` trace observed 512 KiB and 2 MiB stacks, which line up exactly with `COORDINATION_STACK_SIZE` / `AUXILIARY_LOOP_STACK_SIZE` (512 KiB) and `HEAVY_RUNTIME_STACK_SIZE` (2 MiB) in `thread_stacks.zig`.

16 MiB is the value verified on the affected hardware. 8 MiB is only known to be *sufficient on x86_64* (that is the main-thread size), so it is not a safe target for the platform that actually fails.

**This is address space, not resident memory.** Linux commits stack pages on first touch, so a thread that stays shallow still costs a few pages of RSS. The `~1 MB peak RSS` constraint in `CLAUDE.md` is unaffected; virtual size grows, RSS does not.

## Radius

One constant, but it covers every thread that executes a turn:

- the five channel polling loops (`spawnTelegramPolling` and siblings)
- the parallel per-session message workers (`messageTaskWorker`)
- the gateway turn thread
- `inboundDispatcherThread` in `daemon.zig` - which is literally the "inbound worker" named in the issue

The alias to `HEAVY_RUNTIME_STACK_SIZE` is also dropped: websocket gateway loops and subagents have genuinely different depth requirements from the turn path, and tying them together is what let this regress silently.

## Tests

- `session turn stack size can spawn a thread` - matches the existing pattern for the other budgets, and guards against a size the pthread backend would reject.
- `session turn stack clears the main-thread budget` - asserts the budget exceeds both the 8 MiB main-thread reference and `HEAVY_RUNTIME_STACK_SIZE`, so re-aliasing the constant fails the build.

Sanity check: restoring `= HEAVY_RUNTIME_STACK_SIZE` makes the second test fail; restoring the fix makes the suite green again.

## Verification

```
zig build test --summary all   # 7275 pass, 14 skip, 0 fail
```

I do not have aarch64 hardware, so I could not reproduce the segfault or measure the true peak stack usage of the turn path. This change follows the reporter's diagnosis and their verified 16 MiB workaround. Confirmation from anyone running a Raspberry Pi or other aarch64 host would be worth having before release.

## Related

Two commenters on the issue attributed this to the *default* Zig stack size (~512 KiB) in `channel_loop.zig` / `telegram.zig`. That is not what the code does - the size is set explicitly at every spawn site through `thread_stacks.zig`, and the threads on the turn path were getting 2 MiB.

### DIFF
--- a/src/thread_stacks.zig
+++ b/src/thread_stacks.zig
@@ -29,8 +29,17 @@ pub const DAEMON_SERVICE_STACK_SIZE: usize = 1024 * 1024;
 pub const HEAVY_RUNTIME_STACK_SIZE: usize = 2 * 1024 * 1024;
 
 /// Dedicated threads that execute `SessionManager.processMessage*()` /
-/// `Agent.turn()`. Keep this aligned with the heavy runtime budget.
-pub const SESSION_TURN_STACK_SIZE: usize = HEAVY_RUNTIME_STACK_SIZE;
+/// `Agent.turn()`.
+///
+/// The turn path (channel decode -> session -> agent -> provider -> tools) is
+/// the deepest stack in the runtime, and aarch64 frames are larger than x86_64
+/// ones, so 2 MiB overflowed into the guard page and killed the process on
+/// every inbound message there. `nullclaw agent` never hit it because it runs
+/// on the 8 MiB main-thread stack.
+///
+/// This is address space, not resident memory: Linux commits stack pages on
+/// first touch, so a thread that stays shallow still costs a few pages of RSS.
+pub const SESSION_TURN_STACK_SIZE: usize = 16 * 1024 * 1024;
 
 test "coordination stack size can spawn a thread" {
     const thread = try std.Thread.spawn(.{ .stack_size = COORDINATION_STACK_SIZE }, struct {
@@ -44,4 +53,21 @@ test "auxiliary stack size can spawn a thread" {
         fn run() void {}
     }.run, .{});
     thread.join();
+}
+
+test "session turn stack size can spawn a thread" {
+    const thread = try std.Thread.spawn(.{ .stack_size = SESSION_TURN_STACK_SIZE }, struct {
+        fn run() void {}
+    }.run, .{});
+    thread.join();
+}
+
+test "session turn stack clears the main-thread budget" {
+    // Regression: every inbound message segfaulted on aarch64 because the turn
+    // path ran on a 2 MiB thread and overflowed into the guard page, while the
+    // same path was fine on the 8 MiB main-thread stack. Anything at or below
+    // that 8 MiB reference is not a safe budget for the deepest path we have.
+    const main_thread_reference: usize = 8 * 1024 * 1024;
+    try std.testing.expect(SESSION_TURN_STACK_SIZE > main_thread_reference);
+    try std.testing.expect(SESSION_TURN_STACK_SIZE > HEAVY_RUNTIME_STACK_SIZE);
 }


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork at `415b54c8` — applied verbatim including the stack-budget tests. That fork is deployed on 24/7 aarch64/riscv64 single-board hosts, where the 2 MiB turn stack was a live crash. Thank you!

vernonstinebaker (approved):
Approving as maintainer. Validation evidence: this exact patch was applied to the vernonstinebaker/nullclaw fleet line (415b54c8, 2026-09-23) with RED-first testing — the new stack-budget test failed against the 2 MiB alias, passed at 16 MiB; full suite 7535 passed / 9 skipped. It has run in production on four hosts (macOS arm64, riscv64, aarch64 including the affected Radxa-class path) since 2026-09-24. Closes #976 — the reported per-inbound-message SIGSEGV on aarch64 is the exact overflow class this fixes.

PR #986, opened 2026-08-14
GEN-548: make SQLite memory database path configurable

gently-whitesnow:
## Summary

- add `memory.database_path` for SQLite-backed primary memory engines
- keep the existing `<workspace>/memory.db` location when the setting is empty
- resolve relative values from the workspace and accept absolute paths for read-only workspace deployments
- document the setting in the example config and both beginner guides

## Validation

- `zig fmt --check src/`

## Risk and compatibility

The default is unchanged. The setting affects only the primary SQLite-backed memory database; response-cache and vector-sidecar paths retain their existing behavior.


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork at `6316a2c8` — applied verbatim except the one-line `MemoryConfig` field insertion, hand-reconciled next to config fields added since this PR was opened (comment only; not attempting to re-open). Tests and docs included. Read-only-workspace deployments need this.

vernonstinebaker (approved):
Approved. `memory.database_path` keeps the workspace default when empty, resolves relative paths from the workspace, and allows an absolute path. Config round-trip, path resolution, and runtime init are covered, and CI is green on linux, macOS, and Windows.

PR #987, opened 2026-08-15
feat(agent): loop hygiene for long local tool-heavy runs

vernonstinebaker:
## Summary

- Split system prompt into cache-friendly stable prefix + variable datetime tail (`buildStablePrefix`, `buildVariableTail`, `stablePrefixHash`).
- Compress tool outputs before history injection (`result_compress.zig`); observer logs still see full output.
- Add per-turn identical-call loop guard with warn / veto / force-reply (`loop_guard.zig`).
- Honor `agent.parallel_tools` for read-only batches with slot-ordered parallel execution capped by `agent.local_loop.max_parallel_readonly`.
- Add `agent.local_loop` config bundle (parse, save, round-trip test) and document in `docs/en/configuration.md` + `docs/zh/configuration.md`.

Defaults preserve existing behavior when `local_loop` is omitted and `parallel_tools` is false.

## Validation

- `zig fmt --check src/`
- `zig build test --summary all` → 7372/7381 passed (9 skipped), 0 leaks
- Fleet deploy smoke (personal ops hosts):
  - localhost (macOS arm64): `curl http://127.0.0.1:5111/health` → `{"status":"ok"}`
  - OrangePi (riscv64): health OK, `systemctl --user is-active nullclaw` → active
  - Radxa (aarch64): health OK, `systemctl --user is-active nullclaw` → active

## Notes

- Phase 0 only — no constrained-decoding envelope (Phase 1 gated on #971).
- Parallel path mutex-protects policy/rate-limit only; tool I/O runs concurrently for allowlisted read-only batches.
- Same-batch signature dedup still matches sequential semantics when calls are processed in order; duplicate signatures queued together before any execute may run more than once (edge case).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -986,6 +986,36 @@ Tunnel providers for exposing the gateway to the public internet. Required for w
 - `tailscale.auth_key` is optional. Use it when the machine should auto-run `tailscale up` before starting `serve`/`funnel`.
 - Tunnel secrets such as `cloudflare.token`, `ngrok.auth_token`, and `tailscale.auth_key` are encrypted at rest when `secrets.encrypt = true`.
 
+### `agent`
+
+Loop and context hygiene settings for long tool-heavy runs (especially local models).
+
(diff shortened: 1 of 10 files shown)


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork main (`bac43959`) — deployed in production from that fork. Keeping the PR open in case upstream revives.

PR #989, opened 2026-08-19
fix: restore broken star history chart

FaintFlower:
The star history chart on the README is currently broken because it depends on the GitHub stargazer API, which is subject to access restrictions. This change points the chart at star-history.dera.page, a reliable token-free alternative, so the chart renders correctly again.

The chart URL was verified against the repository's remote (nullclaw/nullclaw), so no owner update was needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -968,4 +968,4 @@ MIT — see [LICENSE](LICENSE)
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


vernonstinebaker:
Closing this. The chart on `main` is rendering again.

`README.md` still points at `https://api.star-history.com/svg?repos=nullclaw/nullclaw&type=date&legend=top-left`. Fetched just now, that URL returns HTTP 200 and an SVG of about 65 KB. The `star-history.dera.page` URL in this PR also returns an SVG, so the swap is not required to restore the chart, and it would move the badge to a different host.

Thanks for catching it when the official chart was broken. If `api.star-history.com` fails again, a fresh PR is welcome.

PR #996, opened 2026-09-06
fix(mcp): bound stdio response waits

### LINKED ISSUE #991: MCP stdio calls can hang indefinitely behind the Proxmox launcher lock
## Summary

A standalone `nullclaw agent` invocation can hang indefinitely when the configured stdio MCP server is already owned by the long-lived gateway.

## Reproduction

Environment: Proxmox CT 151, NullClaw 2026.8.22, read-only Proxmox MCP bridge (148 tools).

1. Start `nullclaw-gateway.service`.
2. Run `nullclaw agent --agent proxmox-admin -m "Reply with exactly OK and do not use tools."`.
3. The CLI agent initializes configured MCP tools before processing the prompt.
4. The Proxmox launcher runs `flock /run/lock/proxmox-mcp.lock ssh ...`.
5. The gateway already holds that lock, so the CLI child waits in `flock`.
6. NullClaw stdio MCP `readLine()` has no deadline; the configured/default MCP timeout is not applied to stdio reads.
7. The observed process remains blocked until the external launcher timeout (300s), with no LLM response.

Direct authenticated ClawProxy HTTP from the same CT returned HTTP 200 with `OK`, so the upstream model route was healthy.

## Evidence

- CT gateway remained active with `NRestarts=0` and `/health` 200.
- Hung process tree showed `nullclaw agent -> timeout -> flock -> ssh`.
- The MCP server in CT 153 was healthy and reported 148 tools.
- Source review: `McpServer.sendRequest()` uses blocking stdio `readLine()`; timeout handling is implemented for HTTP/curl transport but not stdio transport.
- The launcher uses a global blocking lock, causing queued duplicate MCP processes.

## Proposed fix

- Apply a real deadline to stdio MCP reads for initialize, tools/list, and tools/call, using `McpServerConfig.timeout_ms`.
- Return a typed timeout/lock-busy error and cleanly terminate the child on handshake/tool failure.
- Change the Proxmox launcher to fail fast on a held lock, or provide one shared MCP connection/broker for gateway and CLI callers.
- Add deterministic tests for stdio timeout and cleanup, plus an integration smoke that verifies a concurrent second caller fails quickly.
- Preserve read-only MCP mode and avoid changing Proxmox permissions or lifecycle behavior.

## Acceptance criteria

- A blocked stdio MCP server returns within the configured timeout.
- No orphaned `ssh`, `flock`, or MCP child processes remain after timeout/cancellation.
- A second caller receives a bounded, actionable error instead of waiting 300s.
- Existing MCP handshake/tool-call tests and the full relevant test suite pass.

be-student:
Fixes #991.

Applies `timeout_ms` to stdio MCP response reads and terminates the server's process group when a request times out. Failed initialization now also cleans up the spawned child.

Validation:
- `zig build test --summary all` (7,373 passed, 9 skipped)
- `zig build -Dtarget=x86_64-windows-gnu`

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,7 +481,8 @@ Config: `~/.nullclaw/config.json` (created by `onboard`)
     "filesystem": {
       "transport": "stdio",
       "command": "npx",
-      "args": ["-y", "@modelcontextprotocol/server-filesystem"]
+      "args": ["-y", "@modelcontextprotocol/server-filesystem"],
+      "timeout_ms": 10000
     },
     "remote": {
       "transport": "http",
@@ -539,6 +540,8 @@ Config: `~/.nullclaw/config.json` (created by `onboard`)
 Config values are literal. NullClaw does not expand `${VAR}` inside `config.json`
 strings, including custom header values. If you need environment-based secrets,
 render `config.json` ahead of time with your own deployment tooling.
+MCP `timeout_ms` bounds each HTTP or stdio request; a timed-out stdio server is
+terminated with its launcher process group so blocked child processes do not remain.
 
 Telegram forum topics:
 
--- a/src/config_types.zig
+++ b/src/config_types.zig
@@ -1880,7 +1880,7 @@ pub const McpServerConfig = struct {
     transport: []const u8 = DEFAULT_TRANSPORT,
     command: []const u8 = "",
     url: ?[]const u8 = null,
-    /// Per-request wall clock timeout for HTTP transport. 0 = default.
+    /// Per-request wall clock timeout for HTTP and stdio transports.
     timeout_ms: u32 = 10_000,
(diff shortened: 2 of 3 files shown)


vernonstinebaker:
Merged into the `vernonstinebaker/nullclaw` fork at `8662764b` — applied verbatim including the blocked-child timeout test (comment only; not attempting to re-open). This directly hardens the stdio MCP servers in production there, where a hung child previously hung the agent. Also noted on issue #991.

vernonstinebaker (approved):
Stdio MCP requests now use `timeout_ms` and the timed-out child is terminated. The regression test covers a server that never replies. CI is green on linux-x86_64, macos-aarch64, and windows-x86_64.

